PR #140, opened 2015-04-28
INTEXT-149 : Provide logic for the HazelcastCacheWritingMessageHandler to accept single entry

erenavsarogullari:
**JIRA :** https://jira.spring.io/browse/INTEXT-149

**HazelcastCacheWritingMessageHandler** has been refactored in order to support single entry and ITopic, MultiMap and ReplicatedMap.

Also SI version has been upgraded to v4.1.3.

### DIFF
--- a/spring-integration-hazelcast/src/main/java/org/springframework/integration/hazelcast/config/xml/HazelcastOutboundChannelAdapterParser.java
+++ b/spring-integration-hazelcast/src/main/java/org/springframework/integration/hazelcast/config/xml/HazelcastOutboundChannelAdapterParser.java
@@ -22,12 +22,12 @@ import org.springframework.beans.factory.support.AbstractBeanDefinition;
 import org.springframework.beans.factory.support.BeanDefinitionBuilder;
 import org.springframework.beans.factory.xml.ParserContext;
 import org.springframework.integration.config.xml.AbstractOutboundChannelAdapterParser;
+import org.springframework.integration.config.xml.IntegrationNamespaceUtils;
 import org.springframework.integration.hazelcast.outbound.HazelcastCacheWritingMessageHandler;
 
-import reactor.util.StringUtils;
-
 /**
- * Hazelcast Outbound Channel Adapter Parser for {@code <int-hazelcast:inbound-channel-adapter />}.
+ * Hazelcast Outbound Channel Adapter Parser for
+ * {@code <int-hazelcast:inbound-channel-adapter />}.
  *
  * @author Eren Avsarogullari
  * @since 1.0.0
@@ -36,16 +36,27 @@ public class HazelcastOutboundChannelAdapterParser extends AbstractOutboundChann
 
 	private static final String CACHE_ATTRIBUTE = "cache";
 
+	private static final String CACHE_EXPRESSION_ATTRIBUTE = "cache-expression";
+
+	private static final String KEY_EXPRESSION_ATTRIBUTE = "key-expression";
+
+	private static final String EXTRACT_PAYLOAD_ATTRIBUTE = "extract-payload";
+
+	private static final String DISTRIBUTED_OBJECT = "distributedObject";
+
 	@Override
 	protected AbstractBeanDefinition parseConsumer(Element element, ParserContext parserContext) {
 		BeanDefinitionBuilder builder = BeanDefinitionBuilder
 				.genericBeanDefinition(HazelcastCacheWritingMessageHandler.class);
 
-		if (!StringUtils.hasText(element.getAttribute(CACHE_ATTRIBUTE))) {
-			parserContext.getReaderContext().error("'" + CACHE_ATTRIBUTE + "' attribute is required.", element);
-		}
-
-		builder.addConstructorArgReference(element.getAttribute(CACHE_ATTRIBUTE));
+		IntegrationNamespaceUtils.setReferenceIfAttributeDefined(builder, element,
+				CACHE_ATTRIBUTE, DISTRIBUTED_OBJECT);
+		IntegrationNamespaceUtils.setValueIfAttributeDefined(builder, element,
+				CACHE_EXPRESSION_ATTRIBUTE);
+		IntegrationNamespaceUtils.setValueIfAttributeDefined(builder, element,
+				KEY_EXPRESSION_ATTRIBUTE);
+		IntegrationNamespaceUtils.setValueIfAttributeDefined(builder, element,
+				EXTRACT_PAYLOAD_ATTRIBUTE);
 
 		return builder.getBeanDefinition();
 	}
--- a/spring-integration-hazelcast/src/main/java/org/springframework/integration/hazelcast/config/xml/HazelcastOutboundChannelAdapterParser.java
+++ b/spring-integration-hazelcast/src/main/java/org/springframework/integration/hazelcast/config/xml/HazelcastOutboundChannelAdapterParser.java
@@ -18,6 +18,7 @@ package org.springframework.integration.hazelcast.config.xml;
 
 import org.w3c.dom.Element;
 
+import org.springframework.beans.factory.config.BeanDefinition;
 import org.springframework.beans.factory.support.AbstractBeanDefinition;
 import org.springframework.beans.factory.support.BeanDefinitionBuilder;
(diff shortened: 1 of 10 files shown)


artembilan:
Hello, @erenavsarogullari !

Sorry for delay and thank you for this PR.

Let me start review your PR!

I wonder how it will be better for you: to address them yourself or receive a polishing commit from me.

spring-integration-hazelcast/src/main/java/org/springframework/integration/hazelcast/config/xml/HazelcastOutboundChannelAdapterParser.java
```diff
@@ -36,16 +36,27 @@
...
-			parserContext.getReaderContext().error("'" + CACHE_ATTRIBUTE + "' attribute is required.", element);
-		}
-
-		builder.addConstructorArgReference(element.getAttribute(CACHE_ATTRIBUTE));
+		IntegrationNamespaceUtils.setReferenceIfAttributeDefined(builder, element,
+				CACHE_ATTRIBUTE, DISTRIBUTED_OBJECT);
+		IntegrationNamespaceUtils.setValueIfAttributeDefined(builder, element,
+				CACHE_EXPRESSION_ATTRIBUTE);
+		IntegrationNamespaceUtils.setValueIfAttributeDefined(builder, element,
+				KEY_EXPRESSION_ATTRIBUTE);
```
artembilan: Typically we do this for any expression attributes:
````java
IntegrationNamespaceUtils.createExpressionDefIfAttributeDefined(CACHE_EXPRESSION_ATTRIBUTE, element);
````
See its source code and my comments in the `HazelcastCacheWritingMessageHandler`

File: spring-integration-hazelcast/src/main/java/org/springframework/integration/hazelcast/outbound/HazelcastCacheWritingMessageHandler.java [hunk 2]
```diff
@@ -16,64 +16,147 @@
...
 
-	public HazelcastCacheWritingMessageHandler(DistributedObject distributedObject) {
+	private boolean extractPayload = true;
+
+	public void setDistributedObject(DistributedObject distributedObject) {
 		Assert.notNull(distributedObject, "'distributedObject' must not be null");
 		this.distributedObject = distributedObject;
 	}
 
+	public void setCacheExpression(String cacheExpression) {
```
artembilan: No, according to the change to the `HazelcastOutboundChannelAdapterParser` this option must be of `Expression` type. As well as the next one.
It's really enough hard operation to parse expression at runtime.

spring-integration-hazelcast/src/main/java/org/springframework/integration/hazelcast/outbound/HazelcastCacheWritingMessageHandler.java
```diff
@@ -16,64 +16,147 @@
...
 	@Override
-	protected void handleMessageInternal(Message<?> message) throws Exception {
+	protected void handleMessageInternal(final Message<?> message) throws Exception {
 		writeToCache(message);
 	}
 
 	@SuppressWarnings({ "unchecked", "rawtypes" })
-	private void writeToCache(Message<?> message) {
-		if (this.distributedObject instanceof IMap) {
-			((IMap<?, ?>) this.distributedObject).putAll((Map) message.getPayload());
```
artembilan: I think it is some regression: we really should still support `put(add)All` operation, if `payload` is of the appropriate type for `distributedObject`.

spring-integration-hazelcast/src/main/java/org/springframework/integration/hazelcast/outbound/HazelcastCacheWritingMessageHandler.java
```diff
@@ -16,64 +16,147 @@
...
+			throw new IllegalStateException("One of 'cache', 'cache-expression' and "
+					+ HazelcastHeaders.CACHE_NAME
+					+ " must be set for cache object definition.");
+		}
+	}
+
+	private DistributedObject parseCacheExpression(final Message<?> message) {
+		Object cacheName = parseExpression(this.cacheExpression, message);
+		return this.getBeanFactory().getBean(cacheName.toString(),
+				DistributedObject.class);
```
artembilan: I don't think so. It's up to end-user to specify some kind of SpEL which must be evaluated to the `DistributedObject`.
We really may not have it as a bean in the context, but `HazelcastInstance.getMap()` and similar may be get into account.
So, it's up to the target application to specify it like this:
````xml
cache-expression="@distributedList"
````
or some other way.
It is more flexible than just rely on the `BeanFactory` state.

File: spring-integration-hazelcast/src/main/java/org/springframework/integration/hazelcast/outbound/HazelcastCacheWritingMessageHandler.java [hunk 5]
```diff
@@ -16,64 +16,147 @@
...
+		else {
+			throw new IllegalStateException(
+					"'key-expression' must be set for IMap, MultiMap and ReplicatedMap");
+		}
+	}
+
+	private Object parseExpression(final String expressionString, final Message<?> message) {
+		Expression expression = new SpelExpressionParser()
+				.parseExpression(expressionString);
+		return expression.getValue(new StandardEvaluationContext(message));
```
artembilan: Your `HazelcastCacheWritingMessageHandler` must `implements IntegrationEvaluationContextAware` and with that the expression evaluation will be like this:
````java
return this.cacheExpression.getValue(this.evaluationContext, message, DistributedObject.class);
````
All expression evaluation must rely on the `this.evaluationContext` to supply `BeanFactoryAccessor` for that `@` bean reference operator and the gain from the `ConversionService` and `PropertyAccessor`s.

File: spring-integration-hazelcast/src/main/java/org/springframework/integration/hazelcast/outbound/HazelcastCacheWritingMessageHandler.java [hunk 6]
```diff
@@ -16,64 +16,147 @@
...
+	}
+
+	private Object getPayloadOrMessage(final Message<?> message) {
```
artembilan: Not sure in this method. We can just extract `objectToStore` variable in the beginning of `handleMessageInternal`  and do the rest of logic.

artembilan:
That's all for now.
Well done!

spring-integration-hazelcast/src/main/java/org/springframework/integration/hazelcast/outbound/HazelcastCacheWritingMessageHandler.java
erenavsarogullari: I have the following concern :

I have implemented like this because the messages may need to be written to different cache objects by deciding at runtime. 

For example : ```cache-expression="headers['CACHE_HEADER']"```.
first message => distributedMap
second message => distributedList

However, as far as i see, we can not support it via ```cache-expression="@distributedList"``` and if the bean does not exist in context.

erenavsarogullari:
Hi @artembilan,

Thanks for the comments.

I have just fixed almost all of them and have a concern. After then, i will push.

erenavsarogullari:
Pushed.

artembilan:
Hi @erenavsarogullari !

Please, find the polishing to your work: https://github.com/artembilan/spring-integration-extensions/commit/b9e61ac86ade70a96721d4be18ec8fdccc0c7bab

If you won't mind, I'll go ahead and merge. Although we need decide which `DistributedObject` types are allowed to use for the `HazelcastCacheWritingMessageHandler`. I mean we have there a flow with `if... else if`, but we don't have `else`. And that last one must throws some `IllegalStateException` to notify end-user that he isn't right with resolved `DistributedObject`. WDYT ?

erenavsarogullari:
Hi @artembilan,

Thanks for polishing ;) It looks ok for me.

Also i agree ```IllegalStateException``` in method ```handleMessageInternal``` for else case.

artembilan:
Merged as https://github.com/spring-projects/spring-integration-extensions/commit/c6975af37b26dc06ea6f80e008065410c664342c

Thank you very much one more time!
Looking forward for more!

PR #141, opened 2015-04-29
ZookeeperMetadataStore support

mbogoevici:
ZookeeperMetadataStore support

- Introduces ListenableMetadataStore and MetadataStoreListener
- ZookeeperMetadataStore implements ListenableMetadataStore


spring-integration-zookeeper/src/main/java/org/springframework/integration/zookeeper/metadata/ZookeeperMetadataStore.java
```diff
@@ -0,0 +1,381 @@
...
+				return null;
+			}
+			catch (KeeperException.NodeExistsException e) {
+				// so the data actually exists, we can read it
+				try {
+					byte[] bytes = this.client.getData().forPath(getPath(key));
+					return Conversions.bytesToString(bytes, "UTF-8");
+				}
+				catch (Exception exceptionDuringGet) {
+					throw new ZookeeperMetadataStoreException("Exception while creating node with key '" + key + "':", e);
```
garyrussell: We're not strictly creating here so this exception might be misleading.

spring-integration-zookeeper/src/main/java/org/springframework/integration/zookeeper/metadata/ZookeeperMetadataStore.java
```diff
@@ -0,0 +1,381 @@
...
+				if (oldValue.equals(Conversions.bytesToString(bytes, "UTF-8"))) {
+					updateNode(key, newValue, currentStat.getVersion());
+				}
+				return true;
+			}
+			catch (KeeperException.NoNodeException e) {
+				// ignore, the node doesn't exist there's nothing to replace
+				return false;
+			}
+			catch (KeeperException.BadVersionException e) {
```
garyrussell: Just so I understand - using currentStat means we're using optimistic locking - otherwise we'd need a Curator `InterProcessMutex` - right?

spring-integration-zookeeper/src/main/java/org/springframework/integration/zookeeper/metadata/ZookeeperMetadataStore.java
```diff
@@ -0,0 +1,381 @@
...
+	@Override
+	public String get(String key) {
+		Assert.notNull(key, "'key' must not be null.");
+		synchronized (this.updateMap) {
+			ChildData currentData = this.cache.getCurrentData(getPath(key));
+			if (currentData == null) {
+				if (this.updateMap.containsKey(key)) {
+					// we have saved the value, but the cache hasn't updated yet
+					// if the value had changed via replication, we would have been notified by the listener
+					return this.updateMap.get(key).getValue();
```
garyrussell: Isn't there a race condition here? e.g. some other instance deleted the node but we didn't get the notification yet?

mbogoevici: You're right, there can be - it's not a problem wrt regular metadata store usage (where eventual consistency after notification has arrived is acceptable), but using the updateMap data first would allow for consistency between returned data and listener invocations.

spring-integration-zookeeper/src/main/java/org/springframework/integration/zookeeper/metadata/ZookeeperMetadataStore.java
mbogoevici: Correct. In fact, optimistic locking is much better in this case, it guarantees correct behaviour even if changes are made directly in ZK.

spring-integration-zookeeper/src/main/java/org/springframework/integration/zookeeper/metadata/ZookeeperMetadataStore.java
mbogoevici: Correct. I'll change the message.

spring-integration-zookeeper/src/main/java/org/springframework/integration/zookeeper/metadata/ZookeeperMetadataStore.java
mbogoevici: Actually, I think that the current functionality is fine - in the case of a delete, the client will just be behind the updates until the delete notification is processed. There's no inconsistency with listener invocations, as by the time the MetadataStoreListeners are called, both the cache and the updateMap will be updated already. We could check the `updateMap` data first, too, but the current functionality isn't fundamentally incorrect.

spring-integration-zookeeper/build.gradle
```diff
@@ -0,0 +1,232 @@
...
+}
+
+task dist(dependsOn: assemble) {
+	group = 'Distribution'
+	description = 'Builds -dist, -docs and distribution archives.'
+}
+
+task wrapper(type: Wrapper) {
+	description = 'Generates gradlew[.bat] scripts'
+	gradleVersion = '1.12'
```
artembilan: Gradle 2.3 would be better.
Less space on local HDD :smile: 
Will fix on merge

spring-integration-zookeeper/gradle.properties
```diff
@@ -0,0 +1 @@
+version=0.0.1.BUILD-SNAPSHOT
```
artembilan: I think I like this versioning manner. And it would be good for our future, when we will have a wish to release it for XD with some existing state and don't wait the whole integration and impl in SI Core.

spring-integration-zookeeper/src/main/java/org/springframework/integration/zookeeper/metadata/Conversions.java
```diff
@@ -0,0 +1,46 @@
...
+
+import java.io.UnsupportedEncodingException;
+
+/**
+ *
+ * Utilities for null-safe conversions.
+ *
+ * @author Marius Bogoevici
+ */
+public class Conversions {
```
artembilan: `org.springframework.util.StringUtils` candidate?
BTW this package (`org.springframework.integration.zookeeper.metadata`) needs `package-info.java`

spring-integration-zookeeper/src/main/java/org/springframework/integration/zookeeper/metadata/MessageStoreListenerAdapter.java
```diff
@@ -0,0 +1,40 @@
...
+ */
+
+package org.springframework.integration.zookeeper.metadata;
+
+/**
+ * Base implementation for a {@link MetadataStoreListener}. Subclasses may override any of the methods.
+ *
+ * @author Marius Bogoevici
+ */
+public class MessageStoreListenerAdapter implements MetadataStoreListener {
```
artembilan: `abstract` ?

spring-integration-zookeeper/src/main/java/org/springframework/integration/zookeeper/metadata/ZookeeperMetadataStore.java
```diff
@@ -0,0 +1,381 @@
...
+				try {
+					byte[] bytes = this.client.getData().forPath(getPath(key));
+					return Conversions.bytesToString(bytes, "UTF-8");
+				}
+				catch (Exception exceptionDuringGet) {
+					throw new ZookeeperMetadataStoreException("Exception while creating node with key '" + key + "':", e);
+				}
+			}
+			catch (Exception e) {
+				throw new ZookeeperMetadataStoreException("Error while trying to set '" + key + "':", e);
```
artembilan: Well, looks like we should have a generic `MetadataStoreException` instead...
But we can live with this for now as well.

spring-integration-zookeeper/src/main/java/org/springframework/integration/zookeeper/metadata/ZookeeperMetadataStore.java
```diff
@@ -0,0 +1,381 @@
...
+		this.client = client;
+	}
+
+	/**
+	 * Encoding to use when storing data in ZooKeeper
+	 *
+	 * @param encoding encoding as text
+	 */
+	public void setEncoding(String encoding) {
+		Assert.notNull(encoding, "'encoding' must not be null.");
```
artembilan: a/notNull/hasText. 
Otherwise `UnsupportedEncodingException` and `IllegalArgumentException`, respectively.

spring-integration-zookeeper/src/main/java/org/springframework/integration/zookeeper/metadata/ZookeeperMetadataStore.java
```diff
@@ -0,0 +1,381 @@
...
+		this.encoding = encoding;
+	}
+
+	/**
+	 * Root node - store entries are children of this node
+	 *
+	 * @param root encoding as text
+	 */
+	public void setRoot(String root) {
+		Assert.notNull(root, "'root' must not be null.");
```
artembilan: DITO

spring-integration-zookeeper/src/main/java/org/springframework/integration/zookeeper/metadata/ZookeeperMetadataStore.java
```diff
@@ -0,0 +1,381 @@
...
+		Assert.isTrue(root.startsWith("/"), "'root' must start with '/'");
+		Assert.isTrue(!root.endsWith("/"), "'root' must not end with '/'");
```
artembilan: We can be here a bit forgiving and add leading `/` and remove `finishing` one.
WDYT?
At least Spring MVC is enough smart on the matter.

spring-integration-zookeeper/src/main/java/org/springframework/integration/zookeeper/metadata/ZookeeperMetadataStore.java
```diff
@@ -0,0 +1,381 @@
...
+
+	private volatile String root = "/SpringIntegration-MetadataStore";
+
+	private volatile String encoding = "UTF-8";
+
+	private volatile PathChildrenCache cache;
+
+	private volatile boolean running = false;
+
+	private volatile boolean autoStartup = false;
```
artembilan: Why not `true`?
OK, I may understand that later from your code...

spring-integration-zookeeper/src/main/java/org/springframework/integration/zookeeper/metadata/ZookeeperMetadataStore.java
```diff
@@ -0,0 +1,381 @@
...
+	}
+
+	public void setPhase(int phase) {
+		this.phase = phase;
+	}
+
+	@Override
+	public String putIfAbsent(String key, String value) {
+		Assert.notNull(key, "'key' must not be null.");
+		Assert.notNull(value, "'value' must not be null.");
```
artembilan: So, is `Assert.hasText` your mental blocker? Or have you really done that intentionally?

mbogoevici: @artembilan keys and values can be empty strings. See https://github.com/spring-projects/spring-integration/blob/master/spring-integration-redis/src/test/java/org/springframework/integration/redis/metadata/RedisMetadataStoreTests.java after which the tests are modeled.

PS. 'mental blocker' isn't something I would use in a public review. Let's stay professional.

artembilan: Oh! Sorry, @mbogoevici !
Just heard that combination of words, but didn't know its translation...
Got it! Will be more accurate.

spring-integration-zookeeper/src/main/java/org/springframework/integration/zookeeper/metadata/ZookeeperMetadataStore.java
```diff
@@ -0,0 +1,381 @@
...
+			synchronized (this.lifecycleMonitor) {
+				if (this.running) {
+					if (this.cache != null) {
+						CloseableUtils.closeQuietly(this.cache);
+					}
+					this.cache = null;
+					this.running = false;
+				}
+			}
+		}
```
artembilan: `callback` isn't invoked here.

spring-integration-zookeeper/src/main/java/org/springframework/integration/zookeeper/metadata/ZookeeperMetadataStore.java
```diff
@@ -0,0 +1,381 @@
...
+				switch (event.getType()) {
+					case CHILD_ADDED:
+						if (ZookeeperMetadataStore.this.updateMap.containsKey(getKey(event.getData().getPath()))) {
+							if (event.getData().getStat().getVersion() >=
+									ZookeeperMetadataStore.this.updateMap.get(getKey(event.getData().getPath())).getVersion()) {
+								ZookeeperMetadataStore.this.updateMap.remove(event.getData().getPath());
+							}
+						}
+						for (MetadataStoreListener listener : ZookeeperMetadataStore.this.listeners) {
+							listener.onAdd(getKey(event.getData().getPath()), Conversions.bytesToString(event.getData().getData(), encoding));
```
artembilan: We can optimize these lines a bit extracting `key` var for those `getKey(event.getData().getPath())` invocations.

spring-integration-zookeeper/src/test/java/org/springframework/integration/zookeeper/metadata/ZookeeperMetadataStoreTests.java
```diff
@@ -0,0 +1,419 @@
...
+	}
+
+	private void closeClient(CuratorFramework client) {
+		try {
+			client.close();
+		}
+		catch (Exception e) {
+			log.warn("Exception thrown while closing client: ", e);
+		}
+	}
```
artembilan: `CloseableUtils.closeQuietly` instead? Minor, of course, and can be still here.

spring-integration-zookeeper/src/test/java/org/springframework/integration/zookeeper/metadata/ZookeeperMetadataStoreTests.java
```diff
@@ -0,0 +1,419 @@
...
+		this.metadataStore.stop();
+		closeClient(this.client);
+		try {
+			testingServer.stop();
+		}
+		catch (IOException e) {
+			log.warn("Exception thrown while shutting down ZooKeeper: ", e);
+		}
+		testingServer.getTempDirectory().delete();
+	}
```
artembilan: Can't we move all this stuff to the `static` class-level initialization phase to increase tests performance? 
I've done that locally and see a race condition with listeners test. But it is expected, because with that `metadataStore` is shared.
Of course, we can live with those `20 sec` and can revise in the future.
Just in case to share with you a sign...

mbogoevici: Yes, probably it's a good idea. We can create a metadataStore per test, and have them use different roots.

spring-integration-zookeeper/src/test/java/org/springframework/integration/zookeeper/metadata/ZookeeperMetadataStoreTests.java
mbogoevici: Yeah, just like we do in the main code.

spring-integration-zookeeper/src/main/java/org/springframework/integration/zookeeper/metadata/ZookeeperMetadataStore.java
mbogoevici: We can make the code simpler for sure.

spring-integration-zookeeper/src/test/java/org/springframework/integration/zookeeper/metadata/ZookeeperMetadataStoreTests.java
artembilan: Another my idea to have shared `store` between tests, like it is in the real applications.

spring-integration-zookeeper/src/main/java/org/springframework/integration/zookeeper/metadata/ZookeeperMetadataStore.java
mbogoevici: Oops.

spring-integration-zookeeper/src/test/java/org/springframework/integration/zookeeper/metadata/ZookeeperMetadataStoreTests.java
mbogoevici: I disagree. Not sure what that would buy us, but it would certainly create issues because the order of test execution is non-deterministic. There are two tests where sharing is in fact tested.

spring-integration-zookeeper/src/main/java/org/springframework/integration/zookeeper/metadata/ZookeeperMetadataStore.java
mbogoevici: no harm done :)

But the earlier comment remains - empty strings are acceptable.

spring-integration-zookeeper/src/main/java/org/springframework/integration/zookeeper/metadata/ZookeeperMetadataStore.java
mbogoevici: I actually think it's `true` everywhere else, so we can change that for sure.

spring-integration-zookeeper/src/main/java/org/springframework/integration/zookeeper/metadata/ZookeeperMetadataStore.java
mbogoevici: I think we want the starting "/" because it's pretty much standard for ZK configurations. We can be lenient, but this is still ZK-specific. There's no notion of relative paths at this point. My 2 cents. The trailing '/' can be removed transparently.

spring-integration-zookeeper/src/main/java/org/springframework/integration/zookeeper/metadata/ZookeeperMetadataStore.java
mbogoevici: Actually, the content of the root configuration is checked differently, see below.

spring-integration-zookeeper/src/main/java/org/springframework/integration/zookeeper/metadata/ZookeeperMetadataStore.java
mbogoevici: Good point.

spring-integration-zookeeper/src/main/java/org/springframework/integration/zookeeper/metadata/Conversions.java
mbogoevici: Maybe we can move this to StringUtils, although the specific role of this class is to isolate null-safe `byte[]` <-> `String` conversions.

spring-integration-zookeeper/src/main/java/org/springframework/integration/zookeeper/metadata/ZookeeperMetadataStore.java
```diff
@@ -0,0 +1,381 @@
...
+		synchronized (this.updateMap) {
+			try {
+				createNode(key, value);
+				return null;
+			}
+			catch (KeeperException.NodeExistsException e) {
+				// so the data actually exists, we can read it
+				try {
+					byte[] bytes = this.client.getData().forPath(getPath(key));
+					return Conversions.bytesToString(bytes, "UTF-8");
```
artembilan: `this.encoding` isntead

spring-integration-zookeeper/src/main/java/org/springframework/integration/zookeeper/metadata/ZookeeperMetadataStore.java
```diff
@@ -0,0 +1,381 @@
...
+	@Override
+	public boolean replace(String key, String oldValue, String newValue) {
+		Assert.notNull(key, "'key' must not be null.");
+		Assert.notNull(oldValue, "'oldValue' must not be null.");
+		Assert.notNull(newValue, "'newValue' must not be null.");
+		synchronized (this.updateMap) {
+			Stat currentStat = new Stat();
+			try {
+				byte[] bytes = this.client.getData().storingStatIn(currentStat).forPath(getPath(key));
+				if (oldValue.equals(Conversions.bytesToString(bytes, "UTF-8"))) {
```
artembilan: `this.encoding` instead

spring-integration-zookeeper/src/main/java/org/springframework/integration/zookeeper/metadata/ZookeeperMetadataStore.java
```diff
@@ -0,0 +1,381 @@
...
+				}
+			}
+			else {
+				if (this.updateMap.containsKey(key)) {
+					// our version is more recent than the cache
+					if (this.updateMap.get(key).getVersion() >= currentData.getStat().getVersion()) {
+						return this.updateMap.get(key).getValue();
+					}
+				}
+				return Conversions.bytesToString(currentData.getData(), "UTF-8");
```
artembilan: `this.encoding` instead

spring-integration-zookeeper/src/main/java/org/springframework/integration/zookeeper/metadata/ZookeeperMetadataStore.java
```diff
@@ -0,0 +1,381 @@
...
+	@Override
+	public String remove(String key) {
+		Assert.notNull(key, "'key' must not be null.");
+		synchronized (this.updateMap) {
+			try {
+				byte[] bytes = this.client.getData().forPath(getPath(key));
+				this.client.delete().forPath(getPath(key));
+				// we guarantee that the deletion will supersede the existing data
+				this.updateMap.put(key, new LocalChildData(null, Integer.MAX_VALUE));
+				return Conversions.bytesToString(bytes, "UTF-8");
```
artembilan: `this.encoding` instead

artembilan:
That's all from my side.
I got some insight after private discussion with @mbogoevici and now I don't have any objections for the solution.
Just those minor transparent issues, which can be addressed on merge.
Although if @mbogoevici won't mind, he can do that himself and it may be a fact that he will find something else from his deep knowledges on the matter.

@garyrussell, your point ?

garyrussell:
LGTM, with the minor polish you pointed out.

mbogoevici:
Moving this to `spring-integration`

PR #146, opened 2015-06-24
Cassandra: Compatibility with IO-2.0

artembilan:


artembilan:
See https://github.com/spring-projects/spring-integration/pull/1482 for more info.
Feel free to merge this one after that and fresh test against Platform with fresh SI in it to be sure that there is no any warn about deprecations.

artembilan:
Merged after rebase

PR #154, opened 2015-12-23
Intext 215

fbalicchia:
JIRA: https://jira.spring.io/browse/INTEXT-215

* add  XSD for CassandraMessageHandler
* add `CassandraOutboundChannelAdapterParser`
* minor refactoring for `CassandraMessageHandler`
* CassandraMessageHandlerTests doesn't change (thanks for exists)
* add `CassandraOutboundChannelAdapterIntegrationTests` that is the mirror of `CassandraMessageHandlerTests` in declarative mode


artembilan:
Wow! 
That's awesome, @fbalicchia !

I think, I'm starting to take a look into this already right now.

Thank you very much!

spring-integration-cassandra/src/main/java/org/springframework/integration/cassandra/config/xml/CassandraNamespaceHandler.java
```diff
@@ -20,12 +20,13 @@
...
 public class CassandraNamespaceHandler extends AbstractIntegrationNamespaceHandler {
 
 	@Override
 	public void init() {
-
+		registerBeanDefinitionParser("outbound-channel-adapter", new CassandraOutboundChannelAdapterParser());
 	}
 
-}
+}
```
artembilan: Your classical issue with a new line in the end of file :smile:

spring-integration-cassandra/src/main/java/org/springframework/integration/cassandra/config/xml/CassandraOutboundChannelAdapterParser.java
```diff
@@ -0,0 +1,114 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.integration.cassandra.config.xml;
+
+import java.util.List;
+
+import org.apache.commons.lang3.StringUtils;
```
artembilan: We should use `org.springframework.util.StringUtils` and don't pull such a dependency (`commons-lang3`) just only for the `StringUtils`.

spring-integration-cassandra/src/main/java/org/springframework/integration/cassandra/config/xml/CassandraOutboundChannelAdapterParser.java
```diff
@@ -0,0 +1,114 @@
...
+    protected AbstractBeanDefinition parseConsumer(Element element, ParserContext parserContext)
+    {
+        BeanDefinitionBuilder builder = BeanDefinitionBuilder.genericBeanDefinition(CassandraMessageHandler.class);
+        
+        
+        String cassandraTemplate = element.getAttribute("cassandra-template");
+        String operationType = element.getAttribute("operation-type");
+        String writeOptionsInstance = element.getAttribute("write-options");
+        String ingestQuery = element.getAttribute("cql-ingest");
+        String outputChannel = element.getAttribute("output-channel");
```
artembilan: This attribute is a part of `gateway`: should be addressed in the separate parser (`AbstractConsumerEndpointParser` implementation)  and therefore `outbound-gateway` component.
Also, yes: for the same `CassandraMessageHandler`.

spring-integration-cassandra/src/test/resources/log4j.properties
```diff
@@ -4,5 +4,6 @@ log4j.appender.stdout=org.apache.log4j.ConsoleAppender
 log4j.appender.stdout.layout=org.apache.log4j.PatternLayout
 log4j.appender.stdout.layout.ConversionPattern=%d{HH:mm:ss.SSS} %-5p [%t][%c] %m%n
 
-log4j.category.org.springframework.integration=WARN
+log4j.category.org.springframework.integration=TRACE
```
artembilan: Just curious: what forced you to switch to `TRACE` ?

spring-integration-cassandra/src/main/java/org/springframework/integration/cassandra/outbound/CassandraMessageHandler.java
```diff
@@ -69,13 +70,23 @@
...
-	public CassandraMessageHandler(CassandraOperations cassandraTemplate, CassandraMessageHandler.Type queryType) {
-		Assert.notNull(cassandraTemplate, "'cassandraTemplate' must not be null.");
+	public void setQueryType(OperationType queryType) {
+		this.queryType = queryType;
+	}
+
+	public CassandraMessageHandler(CassandraOperations cassandraTemplate,
+			CassandraMessageHandler.OperationType queryType) {
+		Assert.notNull(cassandraTemplate,
+				"'cassandraTemplate' must not be null.");
```
artembilan: To be honest, I'm not sure that you should rewrite the code format so much.
I can guess that you may have 90 line length config, but we with Gary prefer 120.
Sorry, if that was confusing for you.

spring-integration-cassandra/src/main/java/org/springframework/integration/cassandra/outbound/CassandraMessageHandler.java
```diff
@@ -69,13 +70,23 @@
...
-		this(cassandraTemplate, Type.INSERT);
+		this(cassandraTemplate, OperationType.INSERT);
 	}
+	
 
-	public CassandraMessageHandler(CassandraOperations cassandraTemplate, CassandraMessageHandler.Type queryType) {
-		Assert.notNull(cassandraTemplate, "'cassandraTemplate' must not be null.");
+	public void setQueryType(OperationType queryType) {
+		this.queryType = queryType;
+	}
```
artembilan: I don't like this setter. It allows to change the behavior of the component. That is dangerous.
So, let's come back just only to the ctor initialization for this option!

spring-integration-cassandra/src/main/java/org/springframework/integration/cassandra/outbound/CassandraMessageHandler.java
```diff
@@ -175,68 +197,74 @@ protected Object handleRequestMessage(Message<?> requestMessage) {
 
 		Object result = payload;
 
-		Type queryType = this.queryType;
+		OperationType queryType = this.queryType;
```
artembilan: Actually I called it like `Type` because it defines exactly the `CassandraMessageHandler` and its behavior.
Right, it is a operation anyway, but let's leave it as is for now.
The XML option may be called as `mode` (see `<int-jpa:outbound-channel-adapter>`) or `command` (see `<int-ftp:outbound-gateway>`).

I'm not sure yet, what have you changed here in the `CassandraMessageHandler`, but let's revert changes to avoid breaking changes, e.g. for Spring Cloud Data Flow: https://github.com/spring-cloud/spring-cloud-stream-modules/tree/master/cassandra-sink

If you'd like to change anything here, let's discuss that in the separate JIRAs!

spring-integration-cassandra/src/main/java/org/springframework/integration/cassandra/config/xml/CassandraOutboundChannelAdapterParser.java
```diff
@@ -0,0 +1,114 @@
...
+        String query = element.getAttribute("query");
+        String producesReply = element.getAttribute("produces-reply");
```
artembilan: This is an internal state and should be populated here in the parser: `false` for `adapter` and `true` for `gateway`.

spring-integration-cassandra/src/main/resources/org/springframework/integration/config/xml/spring-integration-cassandra-1.0.xsd
```diff
@@ -17,5 +17,139 @@
...
+							Reference to an instance of
+							org.springframework.cassandra.core.WriteOptions]]>
+						</xsd:documentation>
+						<tool:expected-type
+							type="org.springframework.cassandra.core.WriteOptions" />
+					</tool:annotation>
+				</xsd:appinfo>
+			</xsd:annotation>
+		</xsd:attribute>
+		<xsd:attribute name="output-channel" type="xsd:string">
```
artembilan: This isn't common attribute. And in our case it should be declared in the separate `gateway` element.
See `spring-integration-amqp.xsd` for sample.

spring-integration-cassandra/src/main/resources/org/springframework/integration/config/xml/spring-integration-cassandra-1.0.xsd
```diff
@@ -17,5 +17,139 @@
...
+			</xsd:annotation>
+		</xsd:attribute>
+		<xsd:attribute name="query" type="xsd:string">
+			<xsd:annotation>
+				<xsd:documentation>
+					<![CDATA[statement to use in prepared statement]]>
+				</xsd:documentation>
+			</xsd:annotation>
+		</xsd:attribute>
+        <xsd:attribute name="produces-reply" default="false" type="xsd:boolean">
```
artembilan: As I said in the parser, this should not be here as well.

spring-integration-cassandra/src/main/resources/org/springframework/integration/config/xml/spring-integration-cassandra-1.0.xsd
```diff
@@ -17,5 +17,139 @@
...
+			</xsd:annotation>
+		</xsd:attribute>
+		<xsd:attribute name="cql-ingest" type="xsd:string">
+			<xsd:annotation>
+				<xsd:documentation>
+					<![CDATA[csql query to ingest data]]>
+				</xsd:documentation>
+			</xsd:annotation>
+		</xsd:attribute>
+		<xsd:attribute name="query" type="xsd:string">
```
artembilan: You have missed here one more attribute - `statement-expression`.

spring-integration-cassandra/src/test/java/org/springframework/integration/cassandra/config/outbound-adapter-parser.xml
```diff
@@ -0,0 +1,73 @@
...
+    <int-cassandra:outbound-channel-adapter id="outbound3" 
+							cassandra-template="cassandraTemplate" 
+							operation-type="STATEMENT"
+							write-options="writeOptions"
+							query="SELECT * FROM book limit :size"
+                            output-channel="resultChannel"
+                            produces-reply="true"
+							auto-startup="true">
+		    <int-cassandra:parameter-expressions name="author" expression="payload"/>
+		    <int-cassandra:parameter-expressions name="size" expression="headers.limit"/>
```
artembilan: M-m-m. This should not be plural - just only `parameter-expression`.

artembilan: In addition I'd like to see `author` in the `SELECT` to be sure that we are good here.
From other side you should agree with me that `SELECT` sample is exactly for the `gateway` behavior.
Although, I'm sure that we can't distinguish the `SELECT` on the adapter level to disable such kind of behavior. 
The strange result there should be signal for end user.

spring-integration-cassandra/src/test/java/org/springframework/integration/cassandra/test/utils/SICassandraTestUtils.java
```diff
@@ -0,0 +1,51 @@
...
+package org.springframework.integration.cassandra.test.utils;
+
+import java.util.ArrayList;
+import java.util.Date;
+import java.util.List;
+import java.util.UUID;
+
+import org.springframework.integration.cassandra.test.domain.Book;
+
+public class SICassandraTestUtils {
```
artembilan: The `@author` block is in the wrong place here. Should be exactly before class name. (OK before class-level annotations).

From other side I don't like this class name. You do here the `Book` stuff. So, maybe something like `BookFactory` or `BookSampler`, but that definitely isn't `Utils`.
And move it, please, to the `domain` package.

spring-integration-cassandra/src/main/java/org/springframework/integration/cassandra/config/xml/CassandraOutboundChannelAdapterParser.java
```diff
@@ -0,0 +1,114 @@
...
+		}
+		
+		if (StringUtils.isNotEmpty(ingestQuery) && !"INSERT".equalsIgnoreCase(operationType)) {
+			parserContext.getReaderContext().error( "Ingest cql query can be apply only with insert operation",element);
+		}
+		else if (StringUtils.isNotEmpty(ingestQuery)) {
+			builder.addPropertyValue("ingestQuery", ingestQuery);
+		}
+		
+        if (StringUtils.isNotEmpty(query)){
```
artembilan: You still have some issues with formatting, so I'm forced to comment on the matter :smile: 
Minor, of course, but you should be ready do not recognize your code in the `master` after my polishing on merge.
E.g. the `{` should be on the same line as its context and with the whitespace before it.

spring-integration-cassandra/src/main/java/org/springframework/integration/cassandra/config/xml/CassandraOutboundChannelAdapterParser.java
```diff
@@ -0,0 +1,114 @@
...
+
+		if (StringUtils.isEmpty(cassandraTemplate)){
+			parserContext.getReaderContext().error("cassandra-template is empty", element);
+		}
+
+		if (StringUtils.isEmpty(operationType)) {
+			parserContext.getReaderContext().error("operation-type need to be specified", element);
+		}
+		
+		OperationType queryType = CassandraMessageHandler.OperationType.toType(operationType);
```
artembilan: We can't do such a `String -> Object` manual conversion in the parser. 
That's because an attribute value can be in `property-placeholder` style or `SpEL`:
```
operation-type="${my.cassandra.operation:#{myConfguration.cassandraOperationFallback}}"
```

I've just checked it just only with the `builder.addPropertyValue("queryType", operationType);` and PP option. 
So, that `toType()` on the `enum` is redundant. As I said before: should be reverted.

spring-integration-cassandra/src/main/java/org/springframework/integration/cassandra/config/xml/CassandraOutboundChannelAdapterParser.java
```diff
@@ -0,0 +1,114 @@
...
+
+        if (StringUtils.isNotEmpty(producesReply)){
+            builder.addPropertyValue("producesReply", producesReply);
+        }
+		
+		
+		List<Element> parameterExpression = DomUtils.getChildElementsByTagName(element, "parameter-expressions");
+		if (!CollectionUtils.isEmpty(parameterExpression)) {
+					ManagedMap<String, Object> parameterExpressionsMap = new ManagedMap<String, Object>();
+					for (Element parmaterExpressionElement : parameterExpression) {
```
artembilan: typo: par**a**m~~a~~**a**ter

spring-integration-cassandra/src/main/resources/org/springframework/integration/config/xml/spring-integration-cassandra-1.0.xsd
```diff
@@ -1,15 +1,15 @@
...
+	xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:beans="http://www.springframework.org/schema/beans"
+	xmlns:tool="http://www.springframework.org/schema/tool"
+	xmlns:integration="http://www.springframework.org/schema/integration"
+	targetNamespace="http://www.springframework.org/schema/integration/cassandra"
+	elementFormDefault="qualified" attributeFormDefault="unqualified">
+	<xsd:import namespace="http://www.springframework.org/schema/beans" />
+	<xsd:import namespace="http://www.springframework.org/schema/tool" />
 	<xsd:import namespace="http://www.springframework.org/schema/integration"
-				schemaLocation="http://www.springframework.org/schema/integration/spring-integration.xsd"/>
+		schemaLocation="http://www.springframework.org/schema/integration/spring-integration-4.2.xsd" />
```
artembilan: I really prefer to make extension project version free. And rely on that SI-core version which is in classpath.
It isn't problem that you have moved this project to the SI-4.2.3, but that would be great that we will be compatible with `4.3` in the future without re-compiling here.

spring-integration-cassandra/src/main/resources/org/springframework/integration/config/xml/spring-integration-cassandra-1.0.xsd
```diff
@@ -17,5 +17,139 @@
...
+					<tool:annotation kind="direct">
+						<tool:expected-type type="java.lang.String" />
+					</tool:annotation>
+				</xsd:appinfo>
+			</xsd:annotation>
+		</xsd:attribute>
+		<xsd:attribute name="cql-ingest" type="xsd:string">
+			<xsd:annotation>
+				<xsd:documentation>
+					<![CDATA[csql query to ingest data]]>
```
artembilan: Typo: c~~s~~ql

spring-integration-cassandra/src/main/resources/org/springframework/integration/config/xml/spring-integration-cassandra-1.0.xsd
```diff
@@ -17,5 +17,139 @@
...
+				</xsd:appinfo>
+			</xsd:annotation>
+		</xsd:attribute>
+		<xsd:attribute name="operation-type" use="required">
+			<xsd:annotation>
+				<xsd:documentation><![CDATA[
+							Indicates the operation to do on cassandra db]]></xsd:documentation>
+				<xsd:appinfo>
+					<tool:annotation kind="direct">
+						<tool:expected-type type="java.lang.String" />
```
artembilan: This must be something like:
```
<xsd:simpleType>
	<xsd:union memberTypes="cassandraHandlerType xsd:string"/>
</xsd:simpleType>
...
<xsd:simpleType name="cassandraHandlerType">
	<xsd:restriction base="xsd:token">
		<xsd:enumeration value="INSERT" />
		<xsd:enumeration value="UPDATE" />
		<xsd:enumeration value="DELETE" />
		<xsd:enumeration value="STATEMENT" />
	</xsd:restriction>
</xsd:simpleType>
```
The attribute should not be `required` and should have `INSERT` as a `default`.

spring-integration-cassandra/src/main/resources/org/springframework/integration/config/xml/spring-integration-cassandra-1.0.xsd
```diff
@@ -17,5 +17,139 @@
...
+			</xsd:annotation>
+		</xsd:attribute>
+		<xsd:attribute name="output-channel" type="xsd:string">
+			<xsd:annotation>
+				<xsd:appinfo>
+					<tool:annotation kind="ref">
+						<xsd:documentation><![CDATA[
+							Reference to an instance of
+							channel where message can be send]]>
+						</xsd:documentation>
```
artembilan: My IDEA reds the `documentation` for me.
It should be moved outside of `appinfo` to the `xsd:annotation`.

spring-integration-cassandra/src/test/java/org/springframework/integration/cassandra/config/outbound-adapter-parser.xml
artembilan: Sorry, I see my `//TODO https://jira.spring.io/browse/DATACASS-213` in the `CassandraMessageHandlerTests` and see your general comment regarding logic reflection from the Java Config to here for XML one.

spring-integration-cassandra/src/main/resources/org/springframework/integration/config/xml/spring-integration-cassandra-1.0.xsd
```diff
@@ -17,5 +17,139 @@
...
+              </xsd:documentation>
+            </xsd:annotation>
+       </xsd:attribute>
+    
+	</xsd:complexType>
+	
+	<xsd:complexType name="queryParameterType">
+		<xsd:annotation>
+			<xsd:documentation><![CDATA[
+				Expression to be evaluated against the Message to replace a query Parameter {placeholder} with the evaluation result.
```
artembilan: We don't replace `{placeholder}`, we use there standard CQL named parameters discovery.
See the sample:
```
SELECT * FROM book WHERE author = :author limit :size
```

spring-integration-cassandra/src/main/java/org/springframework/integration/cassandra/config/xml/CassandraNamespaceHandler.java
fbalicchia: Opss Done

spring-integration-cassandra/src/main/java/org/springframework/integration/cassandra/config/xml/CassandraOutboundChannelAdapterParser.java
fbalicchia: Remove Depedency

spring-integration-cassandra/src/main/java/org/springframework/integration/cassandra/config/xml/CassandraOutboundChannelAdapterParser.java
fbalicchia: `CassandraNamespaceHandler` is splitted
in two parser channel and gateway sharing the commons part.

spring-integration-cassandra/src/test/resources/log4j.properties
fbalicchia: :-| nothing was misprint

spring-integration-cassandra/src/main/java/org/springframework/integration/cassandra/outbound/CassandraMessageHandler.java
fbalicchia: Done. change config line length to 120

spring-integration-cassandra/src/main/java/org/springframework/integration/cassandra/outbound/CassandraMessageHandler.java
fbalicchia: revert

spring-integration-cassandra/src/main/java/org/springframework/integration/cassandra/outbound/CassandraMessageHandler.java
fbalicchia: Sorry  but I did not know about `Spring Cloud Data Flow`, anyway you right that the change to `CassandraMessageHandler` need to be  addressed in separate jira. The goal of this issue is to support namespace handler and for now I've reverted all change on CassandraMessageHandler to  `2e9247711d72959dc1a3509451bd377e34ec1c29`

spring-integration-cassandra/src/main/java/org/springframework/integration/cassandra/config/xml/CassandraOutboundChannelAdapterParser.java
fbalicchia: done

spring-integration-cassandra/src/main/resources/org/springframework/integration/config/xml/spring-integration-cassandra-1.0.xsd
fbalicchia: done

spring-integration-cassandra/src/main/resources/org/springframework/integration/config/xml/spring-integration-cassandra-1.0.xsd
fbalicchia: done

spring-integration-cassandra/src/main/resources/org/springframework/integration/config/xml/spring-integration-cassandra-1.0.xsd
fbalicchia: Oppssss Yes I forget it ! thanks. Fix it and add two unit test. Thanks again for notes

spring-integration-cassandra/src/test/java/org/springframework/integration/cassandra/test/utils/SICassandraTestUtils.java
fbalicchia: Sometimes ..... ;-)

spring-integration-cassandra/src/main/java/org/springframework/integration/cassandra/config/xml/CassandraOutboundChannelAdapterParser.java
fbalicchia: done

spring-integration-cassandra/src/main/java/org/springframework/integration/cassandra/config/xml/CassandraOutboundChannelAdapterParser.java
fbalicchia: Yes you right

spring-integration-cassandra/src/main/java/org/springframework/integration/cassandra/config/xml/CassandraOutboundChannelAdapterParser.java
fbalicchia: done

spring-integration-cassandra/src/main/resources/org/springframework/integration/config/xml/spring-integration-cassandra-1.0.xsd
fbalicchia: Yes... done

spring-integration-cassandra/src/main/resources/org/springframework/integration/config/xml/spring-integration-cassandra-1.0.xsd
fbalicchia: done

spring-integration-cassandra/src/main/resources/org/springframework/integration/config/xml/spring-integration-cassandra-1.0.xsd
fbalicchia: It's better. done

spring-integration-cassandra/src/main/resources/org/springframework/integration/config/xml/spring-integration-cassandra-1.0.xsd
fbalicchia: more outside

spring-integration-cassandra/src/main/resources/org/springframework/integration/config/xml/spring-integration-cassandra-1.0.xsd
fbalicchia: change

fbalicchia:
Hi @artembilan ,
thanks for your time.. I appreciate. I follow your notes and thanks for suggest.
Thanks again see you

artembilan:
Hi, @fbalicchia !

Thank you for the further process.
Reviewing...

spring-integration-cassandra/src/main/java/org/springframework/integration/cassandra/config/xml/CassandraParserUtils.java
```diff
@@ -0,0 +1,92 @@
...
+			BeanDefinitionBuilder builder) {
+
+		String cassandraTemplate = element.getAttribute("cassandra-template");
+		String operationType = element.getAttribute("type");
+		String writeOptionsInstance = element.getAttribute("write-options");
+		String ingestQuery = element.getAttribute("cql-ingest");
+		String query = element.getAttribute("query");
+
+		if (StringUtils.isEmpty(cassandraTemplate)) {
+			parserContext.getReaderContext().error("cassandra-template is empty", element);
```
artembilan: Would be better to say:
`The 'cassandra-template' is required.`

spring-integration-cassandra/src/main/java/org/springframework/integration/cassandra/config/xml/CassandraParserUtils.java
```diff
@@ -0,0 +1,92 @@
...
+		}
+
+		builder.addConstructorArgValue(new RuntimeBeanReference(cassandraTemplate));
```
artembilan: You can use `addConstructorArgReference()` instead.

spring-integration-cassandra/src/main/java/org/springframework/integration/cassandra/config/xml/CassandraParserUtils.java
```diff
@@ -0,0 +1,92 @@
...
+		if (!StringUtils.isEmpty(operationType)) {
+			builder.addConstructorArgValue(operationType);
+		}
+
+		if (!StringUtils.isEmpty(writeOptionsInstance)) {
+			builder.addPropertyReference("writeOptions", writeOptionsInstance);
+		}
```
artembilan: You can use `IntegrationNamespaceUtils.setReferenceIfAttributeDefined()` instead of this `if` and `addPropertyReference`.

spring-integration-cassandra/src/main/java/org/springframework/integration/cassandra/config/xml/CassandraParserUtils.java
```diff
@@ -0,0 +1,92 @@
...
+
+		if (!StringUtils.isEmpty(ingestQuery) && !"INSERT".equalsIgnoreCase(operationType)) {
```
artembilan: 1. `ingestQuery` has a precedence and `queryType` is overridden in the setter. So, this check is redundant. And that's why I asked before do not have setter for the `queryType`.
2. We can't check the actual attribute value in the parser because we don't have it yet here. And that `operationType` maybe is just the PP or SpEL string. The actual value is extracted somewhere later on the Bean population phase.

Just remove and change the next `ingestQuery` population to the `IntegrationNamespaceUtils.setValueIfAttributeDefined()`!

spring-integration-cassandra/src/main/java/org/springframework/integration/cassandra/config/xml/CassandraParserUtils.java
```diff
@@ -0,0 +1,92 @@
...
+			parserContext.getReaderContext().error("Ingest cql query can be apply only with insert operation", element);
+		}
+		else if (!StringUtils.isEmpty(ingestQuery)) {
+			builder.addPropertyValue("ingestQuery", ingestQuery);
+		}
+
+		if (!StringUtils.isEmpty(query)) {
+			builder.addPropertyValue("query", query);
```
artembilan: The same here for the `IntegrationNamespaceUtils.setValueIfAttributeDefined()`, but `ingestQuery`, `query` and `statementExpression` are mutually exclusive. And we should address the check here in the Parser.

spring-integration-cassandra/src/main/java/org/springframework/integration/cassandra/config/xml/CassandraParserUtils.java
```diff
@@ -0,0 +1,92 @@
...
+
+		List<Element> parameterExpression = DomUtils.getChildElementsByTagName(element, "parameter-expression");
+		if (!CollectionUtils.isEmpty(parameterExpression)) {
+			ManagedMap<String, Object> parameterExpressionsMap = new ManagedMap<String, Object>();
+			for (Element parameterExpressionElement : parameterExpression) {
+				String name = parameterExpressionElement.getAttribute("name");
+				String expression = parameterExpressionElement.getAttribute("expression");
+				BeanDefinitionBuilder factoryBeanBuilder = BeanDefinitionBuilder
+						.genericBeanDefinition(ExpressionFactoryBean.class);
+				factoryBeanBuilder.addConstructorArgValue(expression);
```
artembilan: You can use the same `IntegrationNamespaceUtils.createExpressionDefIfAttributeDefined()` here as well instead.

spring-integration-cassandra/src/main/resources/org/springframework/integration/config/xml/spring-integration-cassandra-1.0.xsd
```diff
@@ -17,5 +17,187 @@
...
+				<xsd:extension base="outboundType">
+					<xsd:choice minOccurs="0" maxOccurs="2">
+					<xsd:element name="parameter-expression" type="queryParameterType" minOccurs="0" maxOccurs="unbounded">
+					<xsd:annotation>
+						<xsd:documentation>
+							Specify an expression for parameter variable placeholder in cql statement.
+						</xsd:documentation>
+					</xsd:annotation>
+				    </xsd:element>
+					<xsd:element name="request-handler-advice-chain" type="integration:handlerAdviceChainType" minOccurs="0" maxOccurs="1" />
```
artembilan: The `ref="integration:poller"` must be here as well.
I'm sure we can move them to that common `outboundType`.

spring-integration-cassandra/src/main/resources/org/springframework/integration/config/xml/spring-integration-cassandra-1.0.xsd
```diff
@@ -17,5 +17,187 @@
...
+							</xsd:documentation>
+							<xsd:appinfo>
+								<tool:annotation kind="ref">
+									<tool:expected-type
+										type="org.springframework.messaging.MessageChannel"/>
+								</tool:annotation>
+							</xsd:appinfo>
+						</xsd:annotation>
+					</xsd:attribute>
+					<xsd:attribute name="input-channel" type="xsd:string">
```
artembilan: The gateway's inbound channel is called `request-channel`.

spring-integration-cassandra/src/main/resources/org/springframework/integration/config/xml/spring-integration-cassandra-1.0.xsd
```diff
@@ -17,5 +17,187 @@
...
+		</xsd:complexType>
+	</xsd:element>
+	
+	<xsd:complexType name="outboundType">
+		<xsd:annotation>
+			<xsd:documentation>
+				Common configuration for cassandra adapters.
+			</xsd:documentation>
+		</xsd:annotation>
+		<xsd:attributeGroup ref="integration:channelAdapterAttributes" />
```
artembilan: Yes, we have to use this common group, but it is exactly for the `channel-adapter` definition. 
Should be moved there.
From other side we should mot forget about `lifecycle` and add `<xsd:attributeGroup ref="integration:smartLifeCycleAttributeGroup"/>` to the gateway anyway.

spring-integration-cassandra/src/main/resources/org/springframework/integration/config/xml/spring-integration-cassandra-1.0.xsd
```diff
@@ -17,5 +17,187 @@
...
+							type="org.springframework.cassandra.core.WriteOptions" />
+					</tool:annotation>
+				</xsd:appinfo>
+			</xsd:annotation>
+		</xsd:attribute>
+		<xsd:attribute name="type" default="INSERT" use="optional">
+			<xsd:annotation>
+				<xsd:documentation>
+				<![CDATA[
+					Indicates the operation to do on cassandra db]]>
```
artembilan: Well, actually it isn't operation on the Cassandra. It's really the style of `CassandraMessageHandler` behavior.
So, I'd really rename it to the `mode` and specify description like:
```
Indicates the `CassandraMessageHandler` behavior. Ignored in case of explicit 'query', 'ingest-query' or `statement-expression`.
```
And yes, let's consider to rename the `queryType` property to the `mode` as well!

spring-integration-cassandra/src/test/java/org/springframework/integration/cassandra/config/outbound-adapter-parser-config.xml
```diff
@@ -0,0 +1,87 @@
...
+							auto-startup="true" />
+
+	<int-cassandra:outbound-channel-adapter id="cassandraMessageHandler3" 
+				  	cassandra-template="cassandraTemplate"
+					cql-ingest="insert into book (isbn, title, author, pages, saleDate, isInStock) values (?, ?, ?, ?, ?, ?)"
+					auto-startup="true" />
+
+	<int-cassandra:outbound-channel-adapter id="cassandraMessageHandler4" 
+							cassandra-template="cassandraTemplate" 
+							statement-expression="T(com.datastax.driver.core.querybuilder.QueryBuilder).truncate('book')"
```
artembilan: Minor, but see comment in the `CassandraMessageHandler`:
```
/*
 * Register the Cassandra Query DSL package so they don't need a FQCN for QueryBuilder, for example.
 */
((StandardTypeLocator) typeLocator).registerImport("com.datastax.driver.core.querybuilder");
```
So, you can omit the package in this expression.

spring-integration-cassandra/src/test/resources/cassandra.properties
```diff
@@ -1,3 +1,3 @@
 cassandra.contactpoints=127.0.0.1
-cassandra.port=9042
-cassandra.keyspace=demo
\ No newline at end of file
+cassandra.port=9043
+cassandra.keyspace=demo
```
artembilan: Isn't related to this PR, but I'm going to revise that all our test stuff and redo it to the proper `@EmbeddedCassandra` with random port.

artembilan:
That's all for this round.
I'm really sure the next your commit will be the last one and we can merge.

fbalicchia:
Hi @artembilan  first of all Wishes for new year.

After fixing all notes, I have upgraded spring-data-cassandra to `1.3.2` and cassandra-unit to `2.0.2.2`
I could not run `@EmbeddedCassandra` with a random port cause at at startup it fails, 
please check [test] (https://github.com/fbalicchia/spring-integration-extensions/tree/improvetest/spring-integration-cassandra/src/test/java/org/springframework/integration/cassandra/config/SpringCQLScriptLoadTest.java). 
The same example on `2.2.2.1` works, than I think that the problem is in cassandra-unit version. 
I could not update extension to that version cause driver version used in spring-data-cassandra 
are different from that user by cassandra-unit >= 2.1.9.2

Any suggest are welcome.

Thanks again for help and notes.

artembilan:
Hello @fbalicchia !

Thank you for update!
Looking at this right now...

spring-integration-cassandra/src/main/resources/org/springframework/integration/config/xml/spring-integration-cassandra-1.0.xsd
artembilan: Haven't you tried to more this definition to the `outboundType` as well ?

spring-integration-cassandra/src/main/resources/org/springframework/integration/config/xml/spring-integration-cassandra-1.0.xsd
```diff
@@ -17,5 +17,200 @@
...
+				</xsd:appinfo>
+			</xsd:annotation>
+		</xsd:attribute>
+		<xsd:attribute name="mode" default="INSERT" use="optional">
+			<xsd:annotation>
+				<xsd:documentation>
+				<![CDATA[
+					Indicates the `CassandraMessageHandler` behavior. Ignored in case of explicit 'query', 'ingest-query' or `statement-expression`.
+And yes, let's consider to rename the queryType property to the mode as well!
+Add a line note
```
artembilan: Interesting text :smile: 
I wonder how ti has appeared here :wink:

spring-integration-cassandra/src/test/java/org/springframework/integration/cassandra/config/CassandraOutboundAdapterParserTests.java
```diff
@@ -0,0 +1,104 @@
...
+
+
+/**
+ * @author Filippo Balicchia
+ *
+ */
+public class CassandraOutboundAdapterParserTests {
+	@Test
+	public void minimalConfig() {
+		ClassPathXmlApplicationContext context = new ClassPathXmlApplicationContext("outbound-adapter-parser.xml",
```
artembilan: Would be better to make this test class based on the `@RunWith(SpringJUnit4ClassRunner.class)` anyway: less context loading - better performance!
Yes, the `outbound-adapter-parser.xml` should be renamed to the `CassandraOutboundAdapterParserTests-context.xml` to be loaded by the Spring Test Framework.

spring-integration-cassandra/src/test/java/org/springframework/integration/cassandra/config/outbound-adapter-parser-config.xml
```diff
@@ -0,0 +1,87 @@
...
+
+	<cassandra:cluster contact-points="${cassandra.contactpoints}" port="${cassandra.port}">
+		<cassandra:keyspace action="CREATE_DROP" name="${cassandra.keyspace}" />
+	</cassandra:cluster>
+
+	<cassandra:session keyspace-name="${cassandra.keyspace}" schema-action="RECREATE"/>
+
+
+	
+	<cassandra:converter />
```
artembilan: M-m-m. Let's distinguish configs for parser tests and for integration tests.
Although yes, you can use some common stuff, like those `<int-cassandra:outbound-channel-adapter>` definitions bellow. But this Cassandra stuff should be `mock` for the parser case.

spring-integration-cassandra/src/test/java/org/springframework/integration/cassandra/config/outbound-adapter-parser-config.xml
```diff
@@ -0,0 +1,87 @@
...
+			<util:constant
+				static-field="org.springframework.cassandra.core.ConsistencyLevel.ONE" />
+		</property>
+		<property name="retryPolicy">
+			<util:constant
+				static-field="org.springframework.cassandra.core.RetryPolicy.DOWNGRADING_CONSISTENCY" />
+		</property>
+	</bean>
+	
+	<bean id="resultChannel" class="org.springframework.integration.channel.QueueChannel"/>
```
artembilan: Minor, of course, but for the more knowledge: Spring Integration provides high-level components for such kind of configuration.
```
<int:channel id="resultChannel">
     <int:queue/>
</int:channel>
```

spring-integration-cassandra/src/test/java/org/springframework/integration/cassandra/config/outbound-adapter-parser-fail.xml
```diff
@@ -0,0 +1,44 @@
...
+
+	<cassandra:mapping />
+	<cassandra:converter />
+	<cassandra:template id="cassandraTemplate" />
+	
+	<int:channel id="input">
+		<int:queue/>
+	</int:channel>
+
+
```
artembilan: All the stuff above is redundant and confusing a bit since we worry in this config only about the failure around `query/ingest-query` exclusivity.

spring-integration-cassandra/build.gradle
```diff
@@ -36,11 +36,11 @@ sourceCompatibility = targetCompatibility = 1.7
...
 ext {
 	cassandraVersion = '2.1.5'
-	cassandraUnitVersion = '2.1.3.1'
+	cassandraUnitVersion = '2.1.9.2'
 	jacocoVersion = '0.7.2.201409121644'
 	slf4jVersion = '1.7.12'
-	springDataCassandraVersion = '1.3.0.RELEASE'
-	springIntegrationVersion = '4.2.0.RELEASE'
+	springDataCassandraVersion = '1.3.2.RELEASE'
+	springIntegrationVersion = '4.2.3.RELEASE'
```
artembilan: Well, I'm reaching successful build with this configuration:
```
	cassandraUnitVersion = '2.2.2.1'
	jacocoVersion = '0.7.2.201409121644'
	slf4jVersion = '1.7.13'
	springDataCassandraVersion = '1.3.2.RELEASE'
	springIntegrationVersion = '4.2.4.RELEASE'
```
And 
```
dependencies {
	compile "org.springframework.integration:spring-integration-core:$springIntegrationVersion"
    compile ("org.springframework.data:spring-data-cassandra:$springDataCassandraVersion") {
		exclude group: 'org.liquibase', module: 'liquibase-core'
	}

	testCompile "org.springframework.integration:spring-integration-test:$springIntegrationVersion"

    testCompile("org.cassandraunit:cassandra-unit-spring:$cassandraUnitVersion")

	testRuntime "org.slf4j:slf4j-log4j12:$slf4jVersion"

	jacoco "org.jacoco:org.jacoco.agent:$jacocoVersion:runtime"
}
```
below.

Yes, there is something like:
```
10:58:06.570 ERROR [OptionalTasks:1][org.apache.cassandra.service.CassandraDaemon] Exception in thread Thread[OptionalTasks:1,5,main]
java.lang.NoSuchMethodError: com.codahale.metrics.Snapshot: method <init>()V not found
	at com.codahale.metrics.UniformSnapshot.<init>(UniformSnapshot.java:39)
	at org.apache.cassandra.metrics.EstimatedHistogramReservoir$HistogramSnapshot.<init>(EstimatedHistogramReservoir.java:77)
```
in the logs for `CassandraMessageHandlerTests`, but it is different story and we can go ahead with this and consider the proper `@EmbeddedCassandra` in the next JIRA.

Let's fix this PR particularly for the Namespace support!

spring-integration-cassandra/src/main/resources/org/springframework/integration/config/xml/spring-integration-cassandra-1.0.xsd
fbalicchia: Hi, no I haven't

spring-integration-cassandra/src/main/resources/org/springframework/integration/config/xml/spring-integration-cassandra-1.0.xsd
fbalicchia: :-1:  :-( sorry

fbalicchia:
Hi @artembilan 
have moved `request-handler-advice-chain` to `outBoundType` and now  for test is used `SpringJUnit4ClassRunner`

I implemented CassandraParserUtilsTest tell me WDYT

Thanks again

spring-integration-cassandra/src/main/java/org/springframework/integration/cassandra/config/xml/CassandraOutboundChannelAdapterParser.java
```diff
@@ -0,0 +1,41 @@
...
+
+/**
+ * @author Filippo Balicchia
+ *
+ */
+public class CassandraOutboundChannelAdapterParser extends AbstractOutboundChannelAdapterParser {
+
+	/**
+	 * {@inheritDoc}
+	 */
```
artembilan: Redundant JavaDoc.
Will remove on merge if only an issue.

spring-integration-cassandra/src/main/java/org/springframework/integration/cassandra/config/xml/CassandraOutboundChannelAdapterParser.java
```diff
@@ -0,0 +1,41 @@
+/*
+ * Copyright 2015 the original author or authors.
```
artembilan: Yeah... We are in a new year already.
Since it isn't in master yet, we should update Copyrights of affected classes to the current year. :smile:

artembilan: I can do that on merge though.

spring-integration-cassandra/src/test/java/org/springframework/integration/cassandra/config/CassandraParserUtilsTest.java
```diff
@@ -0,0 +1,95 @@
...
+import org.junit.Test;
+import org.springframework.beans.factory.config.BeanDefinition;
+import org.springframework.beans.factory.support.RootBeanDefinition;
+import org.springframework.integration.cassandra.config.xml.CassandraParserUtils;
+
+/**
+ * @author Filippo Balicchia
+ */
+
+public class CassandraParserUtilsTest {
```
artembilan: Not sure if we need to test this stuff, but it won't hurt.
Anyway the test classes should be named like `*Tests`, not in singular.
Will rename on merge.

artembilan:
Merged as https://github.com/spring-projects/spring-integration-extensions/commit/f7b31228c2839e2a2f77caf27bcc33c04efa4649

Thank you for the contribution, @fbalicchia !

Feel free to go ahead with upgrade to the latest `CassandraUnit` to have more robust test stuff for us.

PR #159, opened 2016-03-31
Intext 219

englishman:
- add unit test
- fix bug

### DIFF
--- a/spring-integration-zip/src/main/java/org/springframework/integration/zip/splitter/UnZipResultSplitter.java
+++ b/spring-integration-zip/src/main/java/org/springframework/integration/zip/splitter/UnZipResultSplitter.java
@@ -33,7 +33,7 @@ import java.util.Map;
  * @since 1.0
  *
  */
-public class UnZipResultSplitter {
+public final class UnZipResultSplitter {
 
 	public List<Message<Object>> splitUnzippedMap(Message<Map<String, Object>> message) {
 		return createMessages(message.getPayload(), message.getHeaders());
--- a/spring-integration-zip/src/main/java/org/springframework/integration/zip/splitter/UnZipResultSplitter.java
+++ b/spring-integration-zip/src/main/java/org/springframework/integration/zip/splitter/UnZipResultSplitter.java
@@ -16,15 +16,16 @@
 
 package org.springframework.integration.zip.splitter;
 
-import java.util.ArrayList;
-import java.util.List;
-import java.util.Map;
-
 import org.apache.commons.io.FilenameUtils;
 import org.springframework.integration.file.FileHeaders;
 import org.springframework.integration.support.MessageBuilder;
 import org.springframework.integration.zip.ZipHeaders;
 import org.springframework.messaging.Message;
+import org.springframework.messaging.MessageHeaders;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
 
 /**
  *
@@ -34,19 +35,35 @@ import org.springframework.messaging.Message;
  */
 public class UnZipResultSplitter {
 
-	public List<Message<Object>> splitUnzippedMap(Map<String, Object> unzippedEntries) {
+	public List<Message<Object>> splitUnzippedMap(Message<Map<String, Object>> message) {
+		return createMessages(message.getPayload(), message.getHeaders());
+	}
 
-		final List<Message<Object>> messages = new ArrayList<Message<Object>>(unzippedEntries.size());
-
+	private List<Message<Object>> createMessages(Map<String, Object> unzippedEntries, MessageHeaders headers) {
+		List<Message<Object>> messages = createList(unzippedEntries.size());
 		for (Map.Entry<String, Object> entry : unzippedEntries.entrySet()) {
-			final String path = FilenameUtils.getPath(entry.getKey());
-			final String filename = FilenameUtils.getName(entry.getKey());
-			final Message<Object> splitMessage = MessageBuilder.withPayload(entry.getValue())
-					.setHeader(FileHeaders.FILENAME, filename)
-					.setHeader(ZipHeaders.ZIP_ENTRY_PATH, path).build();
-			messages.add(splitMessage);
+			messages.add(createMessage(entry, headers));
 		}
 		return messages;
 	}
 
+	private static List<Message<Object>> createList(int initialCapacity) {
+		return new ArrayList<Message<Object>>(initialCapacity);
+	}
+
+	private static Message<Object> createMessage(Map.Entry<String, Object> entry, Map<String, ?> headers) {
+		return MessageBuilder.withPayload(entry.getValue())
+                .setHeader(FileHeaders.FILENAME, getName(entry))
+                .setHeader(ZipHeaders.ZIP_ENTRY_PATH, getPath(entry))
+				.copyHeaders(headers)
+                .build();
+	}
+
+	private static String getName(Map.Entry<String, Object> entry) {
+		return FilenameUtils.getName(entry.getKey());
+	}
+
+	private static String getPath(Map.Entry<String, Object> entry) {
+		return FilenameUtils.getPath(entry.getKey());
+	}
 }
--- a/spring-integration-zip/src/main/java/org/springframework/integration/zip/splitter/UnZipResultSplitter.java
+++ b/spring-integration-zip/src/main/java/org/springframework/integration/zip/splitter/UnZipResultSplitter.java
@@ -16,24 +16,26 @@
 
 package org.springframework.integration.zip.splitter;
 
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+
 import org.apache.commons.io.FilenameUtils;
+
 import org.springframework.integration.file.FileHeaders;
 import org.springframework.integration.support.MessageBuilder;
 import org.springframework.integration.zip.ZipHeaders;
 import org.springframework.messaging.Message;
 import org.springframework.messaging.MessageHeaders;
 
-import java.util.ArrayList;
-import java.util.List;
-import java.util.Map;
-
 /**
  *
  * @author Gunnar Hillert
+ * @author Andriy Kryvtsun
  * @since 1.0
  *
  */
-public final class UnZipResultSplitter {
+public class UnZipResultSplitter {
 
 	public List<Message<Object>> splitUnzippedMap(Message<Map<String, Object>> message) {
 		return createMessages(message.getPayload(), message.getHeaders());
@@ -53,10 +55,10 @@ public final class UnZipResultSplitter {
 
 	private static Message<Object> createMessage(Map.Entry<String, Object> entry, Map<String, ?> headers) {
 		return MessageBuilder.withPayload(entry.getValue())
-                .setHeader(FileHeaders.FILENAME, getName(entry))
-                .setHeader(ZipHeaders.ZIP_ENTRY_PATH, getPath(entry))
+				.setHeader(FileHeaders.FILENAME, getName(entry))
+				.setHeader(ZipHeaders.ZIP_ENTRY_PATH, getPath(entry))
 				.copyHeaders(headers)
-                .build();
+				.build();
 	}
 
 	private static String getName(Map.Entry<String, Object> entry) {
--- a/spring-integration-zip/src/test/java/org/springframework/integration/zip/splitter/UnZipResultSplitterTest-context.xml
+++ b/spring-integration-zip/src/test/java/org/springframework/integration/zip/splitter/UnZipResultSplitterTest-context.xml
@@ -1,23 +1,22 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <beans xmlns="http://www.springframework.org/schema/beans"
-       xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-       xmlns:int="http://www.springframework.org/schema/integration"
-       xsi:schemaLocation="http://www.springframework.org/schema/beans http://www.springframework.org/schema/beans/spring-beans.xsd
+	   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	   xmlns:int="http://www.springframework.org/schema/integration"
+	   xsi:schemaLocation="http://www.springframework.org/schema/beans
+		http://www.springframework.org/schema/beans/spring-beans.xsd
+		http://www.springframework.org/schema/integration
+		http://www.springframework.org/schema/integration/spring-integration.xsd">
 
+	<int:channel id="input"/>
 
+	<int:chain input-channel="input" output-channel="output">
+		<int:splitter>
+			<bean class="org.springframework.integration.zip.splitter.UnZipResultSplitter"/>
+		</int:splitter>
+	</int:chain>
 
-		http://www.springframework.org/schema/integration http://www.springframework.org/schema/integration/spring-integration.xsd">
-
-    <int:channel id="input"/>
-
-    <int:chain input-channel="input" output-channel="output">
-        <int:splitter>
-            <bean class="org.springframework.integration.zip.splitter.UnZipResultSplitter"/>
-        </int:splitter>
-    </int:chain>
-
-    <int:channel id="output">
-        <int:queue/>
-    </int:channel>
+	<int:channel id="output">
+		<int:queue/>
+	</int:channel>
 
 </beans>
--- a/spring-integration-zip/src/test/java/org/springframework/integration/zip/splitter/UnZipResultSplitterTest-context.xml
+++ b/spring-integration-zip/src/test/java/org/springframework/integration/zip/splitter/UnZipResultSplitterTest-context.xml
@@ -1,0 +1,23 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<beans xmlns="http://www.springframework.org/schema/beans"
+       xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+       xmlns:int="http://www.springframework.org/schema/integration"
+       xsi:schemaLocation="http://www.springframework.org/schema/beans http://www.springframework.org/schema/beans/spring-beans.xsd
+
+
+
+		http://www.springframework.org/schema/integration http://www.springframework.org/schema/integration/spring-integration.xsd">
+
+    <int:channel id="input"/>
+
+    <int:chain input-channel="input" output-channel="output">
+        <int:splitter>
+            <bean class="org.springframework.integration.zip.splitter.UnZipResultSplitter"/>
+        </int:splitter>
+    </int:chain>
+
+    <int:channel id="output">
+        <int:queue/>
+    </int:channel>
+
+</beans>
--- a/spring-integration-zip/src/test/java/org/springframework/integration/zip/splitter/UnZipResultSplitterTest.java
+++ b/spring-integration-zip/src/test/java/org/springframework/integration/zip/splitter/UnZipResultSplitterTest.java
@@ -16,8 +16,14 @@
 
 package org.springframework.integration.zip.splitter;
 
+import static org.junit.Assert.*;
+
+import java.util.HashMap;
+import java.util.Map;
+
 import org.junit.Test;
 import org.junit.runner.RunWith;
+
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.integration.channel.QueueChannel;
 import org.springframework.messaging.Message;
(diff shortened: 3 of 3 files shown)


englishman:
I have signed and agree to the terms of the SpringSource Individual Contributor License Agreement.

spring-integration-zip/src/main/java/org/springframework/integration/zip/splitter/UnZipResultSplitter.java
```diff
@@ -16,15 +16,16 @@
...
 import org.apache.commons.io.FilenameUtils;
 import org.springframework.integration.file.FileHeaders;
 import org.springframework.integration.support.MessageBuilder;
 import org.springframework.integration.zip.ZipHeaders;
 import org.springframework.messaging.Message;
+import org.springframework.messaging.MessageHeaders;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
```
artembilan: Hello @englishman !

Thank you for the contribution!

I will review this in nit-pick manner with the hope to avoid similar comment in the future.
Just for case to save your and our time!

artembilan: We use a bit different `import` order:
imports

You can simply reconfigure your IDE to follow with our rules.
Or just let us know that we will forced to revert such a change on merge each time :wink:

spring-integration-zip/src/main/java/org/springframework/integration/zip/splitter/UnZipResultSplitter.java
```diff
@@ -32,21 +33,37 @@
...
 		return messages;
 	}
 
+	private static List<Message<Object>> createList(int initialCapacity) {
+		return new ArrayList<Message<Object>>(initialCapacity);
+	}
+
+	private static Message<Object> createMessage(Map.Entry<String, Object> entry, Map<String, ?> headers) {
+		return MessageBuilder.withPayload(entry.getValue())
+                .setHeader(FileHeaders.FILENAME, getName(entry))
```
artembilan: Tabs, not whitespaces for indents.
That is also an IDE configuration option.

spring-integration-zip/src/main/java/org/springframework/integration/zip/splitter/UnZipResultSplitter.java
```diff
@@ -32,21 +33,37 @@
...
-					.setHeader(ZipHeaders.ZIP_ENTRY_PATH, path).build();
-			messages.add(splitMessage);
+			messages.add(createMessage(entry, headers));
 		}
 		return messages;
 	}
 
+	private static List<Message<Object>> createList(int initialCapacity) {
+		return new ArrayList<Message<Object>>(initialCapacity);
+	}
```
artembilan: To be honest I don't see reason in this method, because of the single line of code.
Minor of course, but everything must have an explanation.
Would you mind to share your mind on the matter?

spring-integration-zip/src/main/java/org/springframework/integration/zip/splitter/UnZipResultSplitter.java
```diff
@@ -32,21 +33,37 @@
...
+				.copyHeaders(headers)
+                .build();
+	}
+
+	private static String getName(Map.Entry<String, Object> entry) {
+		return FilenameUtils.getName(entry.getKey());
+	}
+
+	private static String getPath(Map.Entry<String, Object> entry) {
+		return FilenameUtils.getPath(entry.getKey());
```
artembilan: The same concern for both these methods, too.
Moreover in the scope of one method the compiler will optimize two `entry.getKey()` to the single one.
But with this style we lose that feature.
So, again I would like to see everything within the `createMessage()` method.

spring-integration-zip/src/test/java/org/springframework/integration/zip/splitter/UnZipResultSplitterTest-context.xml
```diff
@@ -0,0 +1,23 @@
...
+        <int:splitter>
+            <bean class="org.springframework.integration.zip.splitter.UnZipResultSplitter"/>
+        </int:splitter>
+    </int:chain>
+
+    <int:channel id="output">
+        <int:queue/>
+    </int:channel>
+
+</beans>
```
artembilan: Minor, of course, but annoying here on GH, and makes unnecessary changes during review locally, but any file must have a new line in the end.
That's also an IDE configuration option.

spring-integration-zip/src/test/java/org/springframework/integration/zip/splitter/UnZipResultSplitterTest-context.xml
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<beans xmlns="http://www.springframework.org/schema/beans"
+       xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+       xmlns:int="http://www.springframework.org/schema/integration"
+       xsi:schemaLocation="http://www.springframework.org/schema/beans http://www.springframework.org/schema/beans/spring-beans.xsd
+
+
+
+		http://www.springframework.org/schema/integration http://www.springframework.org/schema/integration/spring-integration.xsd">
```
artembilan: Strange formatting, but I'm sure can be fix with the reformat tool in the IDE.
And again tab not whitespaces.
You can find some common Spring Projects IDE configuration options here:
https://github.com/spring-projects/spring-framework/tree/master/src/idea
https://github.com/spring-projects/spring-framework/tree/master/src/eclipse

spring-integration-zip/src/test/java/org/springframework/integration/zip/splitter/UnZipResultSplitterTest.java
```diff
@@ -0,0 +1,76 @@
...
+import org.springframework.test.context.ContextConfiguration;
+import org.springframework.test.context.junit4.SpringJUnit4ClassRunner;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+
+@RunWith(SpringJUnit4ClassRunner.class)
```
artembilan: You can add your name to the `@author` JavaDoc.
Or we do that for you on merge :smile:

spring-integration-zip/src/test/java/org/springframework/integration/zip/splitter/UnZipResultSplitterTest.java
```diff
@@ -0,0 +1,76 @@
...
+@ContextConfiguration("UnZipResultSplitterTest-context.xml")
+public class UnZipResultSplitterTest {
```
artembilan: Our naming convention is like ...Test**s**.
From other hand you don't need to specify the context file for the `@ContextConfiguration` annotation if that file really has a name like `[testClassName]-context.xml` and they both are in the same package.

spring-integration-zip/src/main/java/org/springframework/integration/zip/splitter/UnZipResultSplitter.java
```diff
@@ -32,21 +33,37 @@
  * @since 1.0
  *
  */
-public class UnZipResultSplitter {
-
-	public List<Message<Object>> splitUnzippedMap(Map<String, Object> unzippedEntries) {
+public final class UnZipResultSplitter {
```
artembilan: Not sure that making it as a `final` is a good idea.
Someone may decide to extend the logic but still reuse our solution.

At last it's up to end-user to extend the class on his own faith.
Since this class doesn't provide an access to any critical Framework block, that is really an overhead to protect it with `final`.

artembilan:
That's all. 
I can fix that everything on merge, because the proposal fix meets the requirements, but I would like to hear from you back on the matter. 
And it is really would be better that you address all of them yourself to used to make PRs for our projects.

Thank you for the contribution again!

spring-integration-zip/src/main/java/org/springframework/integration/zip/splitter/UnZipResultSplitter.java
englishman: @artembilan I've used proposed IDE code conventions and updated files. BTW it'll be better to add link to IDE code conventions files into direct into CONTRIBUTING.adoc file.

spring-integration-zip/src/main/java/org/springframework/integration/zip/splitter/UnZipResultSplitter.java
englishman: @artembilan I splitted one big method into some new smaller methods cause the source method broke Single Responsibility Principle. With current method decomposition it's much understandable what's going on in each of them.

spring-integration-zip/src/main/java/org/springframework/integration/zip/splitter/UnZipResultSplitter.java
englishman: @artembilan you shouldn't actually think how compiler do optimizations beforehand. Moreover, usually optimizations make code more tricky and less maintainable. So, IMHO, any optimization should be done 1. if something really works slow and 2. in that place where this slowness is located on a base of objective profiler results. Usually I try to write simple, understandable and maintainable code and only later do some optimization where and when it's really needed.

spring-integration-zip/src/test/java/org/springframework/integration/zip/splitter/UnZipResultSplitterTest-context.xml
englishman: @artembilan why do you need empty line at the and of any file?

spring-integration-zip/src/test/java/org/springframework/integration/zip/splitter/UnZipResultSplitterTest.java
englishman: @artembilan done

spring-integration-zip/src/test/java/org/springframework/integration/zip/splitter/UnZipResultSplitterTest.java
englishman: @artembilan I didn't know such Spring convention. Thanks!

spring-integration-zip/src/main/java/org/springframework/integration/zip/splitter/UnZipResultSplitter.java
englishman: @artembilan removed `final` cause of it may breaks someone's code if it *already* succeeded from `UnZipResultSplitter`.

spring-integration-zip/src/main/java/org/springframework/integration/zip/splitter/UnZipResultSplitter.java
artembilan: OK. Will do if you find it annouying just as comment from here. The code convention is not so big deal as I said and that we request you to follow it just in case to save the time. Otherwise you will see such a comment from us in you PRs and they won't be (maybe) merged until addressing the code format too.

spring-integration-zip/src/main/java/org/springframework/integration/zip/splitter/UnZipResultSplitter.java
artembilan: Well, that is good idea to simplify the code from the level of responsibility, but this creatList() method is really responsible for no more like ArrayList creation. The call for this method introduce more jumps for my mind to follow than its real power.

spring-integration-zip/src/main/java/org/springframework/integration/zip/splitter/UnZipResultSplitter.java
artembilan: Again: any utility method introduction should bring its general purpose. this two methods just do a single particular operators. We should think about optimization ASAP. IMO double getKey() invocation is overhead. Plus I don't see so much gain for this methods. They don't provide comprehensive logic to be severed from the main method. From other hand following with your logic , it doesn't make code understable, at least for me... So, since I have a responsibility to review and merge your code to the public community wide repository, you should convince me in these methods. Sorry that I don't think like you, but it really looks like we are fruits of different fields 😀

spring-integration-zip/src/test/java/org/springframework/integration/zip/splitter/UnZipResultSplitterTest-context.xml
artembilan: That's just one more our code convention . Many codestyle tools require it. Or you follow it or not. No more questions on the matter, please!

spring-integration-zip/src/main/java/org/springframework/integration/zip/splitter/UnZipResultSplitter.java
artembilan: That's not only a reason. Even if it is a new class we should think twice to make it final or package protected. They really may be useful in the future for end user code. Don't forget we develope open source framework, so any code should be available for usage and extension. Only critical infrastructure code may be protected that way.

spring-integration-zip/src/main/java/org/springframework/integration/zip/splitter/UnZipResultSplitter.java
```diff
@@ -21,32 +21,51 @@
 import java.util.Map;
```
artembilan: One mode code changing requirement from us: the Copyring should be updated to the current year - 2015-2016 for this case. Minor, of course, but seems for me we have such a note In the Contribution Guideline

spring-integration-zip/src/test/java/org/springframework/integration/zip/splitter/UnZipResultSplitterTest.java
```diff
@@ -0,0 +1,77 @@
...
+		payload.put("file2", "data2");
+
+		Message<?> inMessage = MessageBuilder.withPayload(payload)
+				.setHeader(headerName, headerValue)
+				.build();
+
+		input.send(inMessage);
+
+		// check message 1
+		checkMessage(output.receive(), headerName, headerValue);
```
artembilan: We should use the receive() with the time (10 sec is our choice) to avoid an infinite looping on the CPU, when our test fails for some reason. Plus I'd like to see that we really receive two different messages - asserts for their payloads.

spring-integration-zip/src/test/java/org/springframework/integration/zip/splitter/UnZipResultSplitterTest.java
```diff
@@ -0,0 +1,77 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.integration.zip.splitter;
+
+import static org.junit.Assert.*;
```
artembilan: We don't like asterisk imports. here (http://stackoverflow.com/questions/147454/why-is-using-a-wild-card-with-a-java-import-statement-bad) you can find a long discussion on the matter, so I don't want to start our own...

spring-integration-zip/src/test/java/org/springframework/integration/zip/splitter/UnZipResultSplitterTest.java
```diff
@@ -0,0 +1,77 @@
...
+		// check message 1
+		checkMessage(output.receive(), headerName, headerValue);
+		// check message 2
+		checkMessage(output.receive(), headerName, headerValue);
+	}
+
+	private static void checkMessage(Message<?> message, String headerName, String headerValue) {
+		assertNotNull(message);
+		MessageHeaders headers = message.getHeaders();
+		assertNotNull(headers);
```
artembilan: Minor, of course, but this assert is redundant. The GenericMessage can't be without headers. Even if it can by some particular Message implementation it's not already this test responsibility . Here we really should just check that we receive the request headers in the splitted messages. Otherwise we are going to end with testing the whole framework and beyond

spring-integration-zip/src/test/java/org/springframework/integration/zip/splitter/UnZipResultSplitterTest.java
```diff
@@ -0,0 +1,77 @@
...
+		assertEquals(headerValue, headers.get(headerName));
+	}
+}
```
artembilan: And the new line in the end should be fixed, please.

spring-integration-zip/src/main/java/org/springframework/integration/zip/splitter/UnZipResultSplitter.java
```diff
@@ -21,32 +21,51 @@
...
+	private static List<Message<Object>> createList(int initialCapacity) {
+		return new ArrayList<Message<Object>>(initialCapacity);
+	}
+
+	private static Message<Object> createMessage(Map.Entry<String, Object> entry, Map<String, ?> headers) {
+		return MessageBuilder.withPayload(entry.getValue())
+				.setHeader(FileHeaders.FILENAME, getName(entry))
+				.setHeader(ZipHeaders.ZIP_ENTRY_PATH, getPath(entry))
+				.copyHeaders(headers)
+				.build();
```
artembilan: @englishman ,

We recently faced a performance issue with the message re-creation in the reply-producing component. The `Splitter` is one of them. So, we switched there to the `MessageBuilder` objects in between request and reply. Like this our `ZipSplitter` particular case.
Would you mind to take a look if we can avoid this `.build()` here as well and just propagate `MessageBuilder` as is to the further processing?

Thanks

spring-integration-zip/src/main/java/org/springframework/integration/zip/splitter/UnZipResultSplitter.java
englishman: @artembilan yes, the purpose of this method is just creation of `List<...>` instance. Moreover, it's possible to make `createList` as `protected` and class user can define it's own `List<...>` implementation without changing the main method in his subclass.

So for me introducing this method is a step forward to make whole class more flexible. But it's another question how this class should be implemented to be reusable, flexible and extendable. And definitely it's not a purpose of my PR. So if you think `createList` isn't needed here - remove it.

spring-integration-zip/src/main/java/org/springframework/integration/zip/splitter/UnZipResultSplitter.java
englishman: @artembilan the purpose of these methods isn't general. It's not utility methods like e.g. `Math#max`. These methods are small atom pieces of algorithm (get file name and get file path from zipped entry). Moreover, again, with adding `protected` and removing `static` you can provide this methods for class successor for modification in the future. E.g. get file name may extract the name in non standard format (not in format provided by `FilenameUtils` by default implementation) or even get file name from entry value.

Again if you think more about [Premature Optimization](http://c2.com/cgi/wiki?PrematureOptimization) than flexibility and maintainability then... well... remove them. And I'll make note about this as the project style in my future code changes :)

spring-integration-zip/src/main/java/org/springframework/integration/zip/splitter/UnZipResultSplitter.java
englishman: @artembilan yes, absolutely agree with you.

But what kind of extension allows the class before my PR? Just simple subclassing? It's useless: subclass can't change any part of logic in solid `splitUnzippedMap` method. Only one way user can 'extend' it: call smth before or/and after `splitUnzippedMap`. But the same thing may be done only with delegation without any subclassing!

My changes provide splitting `splitUnzippedMap` into some pieces of algorithm user can easily and safely change in subclass. In my PR `splitUnzippedMap` is more like template method. That's was idea. Other good things about this splitting: decrease code complexity and increase code cleverness and, thus, maintainability.

But, again, my PR isn't about create better architecture for the class. It's only for bug fix and some obvious 'scout work' and refactoring.

spring-integration-zip/src/main/java/org/springframework/integration/zip/splitter/UnZipResultSplitter.java
artembilan: OK. That is a good catch!

Please, make all these new methods as `protected` and modify them do not deal with `Map.Entry`, but with `zipEntryName` and `payload` as two separate variables and therefore parameters for these methods. E.g.:
````java
protected String getName(String zipEntryName, Object payload) {
	return FilenameUtils.getName(zipEntryName);
}
````

Or... Try to convince me that the `Map.Entry` is a good entity to go :wink: 

Thank you for being so insistent!

spring-integration-zip/src/main/java/org/springframework/integration/zip/splitter/UnZipResultSplitter.java
englishman: @artembilan pls, clarify proposed change affect another methods too. So are you want to change finally signature of the `public` method `List<Message<Object>> splitUnzippedMap(...)` to `List<MessageBuilder<Object>> splitUnzippedMap(...)`?

artembilan: Right, that is purpose: to return a `List<MessageBuilder<Object>>` instead for better performance in the wrapper, which creates messages anyway by its logic.

Since `Message` is immutable, adding any new header to it provides some overhead which we would like to avoid if we can change something in our code.

I may guess that you will start to say that it would be a breaking change, but we are here just only after an `M1` and such a change should be expectable.

From other hand we really can do that in the separate PR, of course. But for now I would like to know if we can do that at all. There might be need changes in some other place.

If you don't like to do that, that's fine, too. I'll take care about that myself after merging your PR.

Anyway I have a fully different idea in mind for the `ZipSplitter` solution. AS you noticed this `UnZipResultSplitter` does not have so much value and may be enough easy re-implement by end-user from zero.

So, the idea is based on the `AbstractMessageSplitter` implementation and so to get a gain from the direct framework code usage.

artembilan:
So, @englishman ,

1. The issue for the redesign https://github.com/spring-projects/spring-integration-extensions/issues/160
2. Please, revert your `static` methods as redundant and just fix the request problem - missed headers in the splitted messages.
3. Please, fix the test-case as I commented. That is really enough critical do not block the test-case infinitely

That is how this PR should look.

Or just let me know that I can do that everything on merge. I'll share the final commit here anyway.

Thank you and looking forward for more and already without so long discussions!

artembilan:
Closed in favor of #161

PR #161, opened 2016-04-04
Intext 219 only bug fix

englishman:
- add unit test
- fix bug

### DIFF
--- a/spring-integration-zip/src/main/java/org/springframework/integration/zip/splitter/UnZipResultSplitter.java
+++ b/spring-integration-zip/src/main/java/org/springframework/integration/zip/splitter/UnZipResultSplitter.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015 the original author or authors.
+ * Copyright 2015-2016 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/spring-integration-zip/src/main/java/org/springframework/integration/zip/splitter/UnZipResultSplitter.java
+++ b/spring-integration-zip/src/main/java/org/springframework/integration/zip/splitter/UnZipResultSplitter.java
@@ -21,20 +21,25 @@ import java.util.List;
 import java.util.Map;
 
 import org.apache.commons.io.FilenameUtils;
+
 import org.springframework.integration.file.FileHeaders;
 import org.springframework.integration.support.MessageBuilder;
 import org.springframework.integration.zip.ZipHeaders;
 import org.springframework.messaging.Message;
+import org.springframework.messaging.MessageHeaders;
 
 /**
  *
  * @author Gunnar Hillert
+ * @author Andriy Kryvtsun
  * @since 1.0
  *
  */
 public class UnZipResultSplitter {
 
-	public List<Message<Object>> splitUnzippedMap(Map<String, Object> unzippedEntries) {
+	public List<Message<Object>> splitUnzippedMap(Message<Map<String, Object>> message) {
+		final MessageHeaders headers = message.getHeaders();
+		final Map<String, Object> unzippedEntries = message.getPayload();
 
 		final List<Message<Object>> messages = new ArrayList<Message<Object>>(unzippedEntries.size());
 
@@ -43,7 +48,9 @@ public class UnZipResultSplitter {
 			final String filename = FilenameUtils.getName(entry.getKey());
 			final Message<Object> splitMessage = MessageBuilder.withPayload(entry.getValue())
 					.setHeader(FileHeaders.FILENAME, filename)
-					.setHeader(ZipHeaders.ZIP_ENTRY_PATH, path).build();
+					.setHeader(ZipHeaders.ZIP_ENTRY_PATH, path)
+					.copyHeaders(headers)
+					.build();
 			messages.add(splitMessage);
 		}
 		return messages;
--- a/spring-integration-zip/src/main/java/org/springframework/integration/zip/splitter/UnZipResultSplitter.java
+++ b/spring-integration-zip/src/main/java/org/springframework/integration/zip/splitter/UnZipResultSplitter.java
@@ -49,7 +49,7 @@ public class UnZipResultSplitter {
 			final Message<Object> splitMessage = MessageBuilder.withPayload(entry.getValue())
 					.setHeader(FileHeaders.FILENAME, filename)
 					.setHeader(ZipHeaders.ZIP_ENTRY_PATH, path)
-					.copyHeaders(headers)
+					.copyHeadersIfAbsent(headers)
 					.build();
 			messages.add(splitMessage);
 		}
--- a/spring-integration-zip/src/test/java/org/springframework/integration/zip/splitter/UnZipResultSplitterTests-context.xml
+++ b/spring-integration-zip/src/test/java/org/springframework/integration/zip/splitter/UnZipResultSplitterTests-context.xml
@@ -1,0 +1,22 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<beans xmlns="http://www.springframework.org/schema/beans"
+	   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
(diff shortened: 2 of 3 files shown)


artembilan:
Thank you, @englishman, for the understanding and so fast fix!

spring-integration-zip/src/main/java/org/springframework/integration/zip/splitter/UnZipResultSplitter.java
```diff
@@ -43,7 +48,9 @@
 			final String filename = FilenameUtils.getName(entry.getKey());
 			final Message<Object> splitMessage = MessageBuilder.withPayload(entry.getValue())
 					.setHeader(FileHeaders.FILENAME, filename)
-					.setHeader(ZipHeaders.ZIP_ENTRY_PATH, path).build();
+					.setHeader(ZipHeaders.ZIP_ENTRY_PATH, path)
+					.copyHeaders(headers)
```
artembilan: I think in this context the `copyHeadersIfAbsent()` would be better.
In case of an upstream message already has the mentioned `FileHeaders.FILENAME` or `ZipHeaders.ZIP_ENTRY_PATH` headers.
If we use this `copyHeaders()`, we override just added headers based on the zip entry. And that is wrong.
I think that should be covered with the test-case as well.

spring-integration-zip/src/test/java/org/springframework/integration/zip/splitter/UnZipResultSplitterTests-context.xml
```diff
@@ -0,0 +1,22 @@
...
+		<int:splitter>
+			<bean class="org.springframework.integration.zip.splitter.UnZipResultSplitter"/>
+		</int:splitter>
+	</int:chain>
+
+	<int:channel id="output">
+		<int:queue/>
+	</int:channel>
+
+</beans>
```
artembilan: The new line in the end of file.

spring-integration-zip/src/test/java/org/springframework/integration/zip/splitter/UnZipResultSplitterTests.java
```diff
@@ -0,0 +1,75 @@
...
+		payload.put("file2", "data2");
+
+		Message<?> inMessage = MessageBuilder.withPayload(payload)
+				.setHeader(headerName, headerValue)
+				.build();
+
+		input.send(inMessage);
+
+		checkMessage(output.receive(TIMEOUT), headerName, headerValue);
+		checkMessage(output.receive(TIMEOUT), headerName, headerValue);
```
artembilan: We should assert the `payload` as well. Just for the future cases when we will change anything and will be surprised that don't receive expected messages.

spring-integration-zip/src/test/java/org/springframework/integration/zip/splitter/UnZipResultSplitterTests-context.xml
```diff
@@ -0,0 +1,22 @@
...
+		http://www.springframework.org/schema/integration
+		http://www.springframework.org/schema/integration/spring-integration.xsd">
+
+	<int:channel id="input"/>
+
+	<int:chain input-channel="input" output-channel="output">
+		<int:splitter>
+			<bean class="org.springframework.integration.zip.splitter.UnZipResultSplitter"/>
+		</int:splitter>
+	</int:chain>
```
artembilan: Minor, but this `<chain>` wrapping is redundant.
Here we just test a `splitter` functionality.

artembilan:
That's all.

I can address them on merge, but would like to hear from you back first of all.

Thank you!

spring-integration-zip/src/test/java/org/springframework/integration/zip/splitter/UnZipResultSplitterTests-context.xml
englishman: @artembilan how splitter can be linked to input and output channels without `chain`?

artembilan: The top-level `<int:splitter>` element has those attributes as well. 
Just move them from the `<chain>` and remove the `<chain>`!

spring-integration-zip/src/test/java/org/springframework/integration/zip/splitter/UnZipResultSplitterTests.java
englishman: @artembilan it's slippery thing: cause of `Map` interface allows usage of implementation without elements order guarantee

spring-integration-zip/src/main/java/org/springframework/integration/zip/splitter/UnZipResultSplitter.java
englishman: @artembilan agreed

spring-integration-zip/src/test/java/org/springframework/integration/zip/splitter/UnZipResultSplitterTests.java
artembilan: So, use `LinkedHashMap` to be sure that iteration in the `UnZipResultSplitter` is done in the proper order!
Since the `output` is `QueueChannel`, the delivery order will be maintained as well.

englishman:
@artembilan i'm done

spring-integration-zip/src/test/java/org/springframework/integration/zip/splitter/UnZipResultSplitterTests-context.xml
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<beans xmlns="http://www.springframework.org/schema/beans"
+	   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	   xmlns:int="http://www.springframework.org/schema/integration"
+	   xsi:schemaLocation="http://www.springframework.org/schema/beans
+		http://www.springframework.org/schema/beans/spring-beans.xsd
+		http://www.springframework.org/schema/integration
+		http://www.springframework.org/schema/integration/spring-integration.xsd">
+
+	<int:channel id="input"/>
```
artembilan: Hello, @englishman !

Glad to see your progress.

I'm merging your PR already.

Right now just for note for the future. The Framework has a feature to create channels automatically just by the name from the `input-channel` of the consumer.
So, we even don't need to declare this element to make it working.

artembilan:
Merged as https://github.com/spring-projects/spring-integration-extensions/commit/8dcca5b676f19dc0e070fd4f986311c70aa11a77.

Thank you for contribution.

Looking forward for more!

spring-integration-zip/src/test/java/org/springframework/integration/zip/splitter/UnZipResultSplitterTests-context.xml
englishman: @artembilan so can I omit `input-channel="input"` at all in the splitter declaration? is this true about `output-channel="output"`?

englishman:
@artembilan interesting... my PR status after your applying is **Closed with unmerged commits**
Why you haven't simply merge PR? It's just curiosity :)

spring-integration-zip/src/test/java/org/springframework/integration/zip/splitter/UnZipResultSplitterTests-context.xml
artembilan: Yes, in most cases we can omit their declaration at all.
But not everywhere yet...
That is a bit spread out task in many places and we appreciate any issue with `auto-create channels`.

artembilan:
We apply only **single** commit into the `master` per PR to supply the linear commit history.

So, I rebased and squashed your PR to the single commit.

Also I made some simple code style polishing, like redundant `final` for local vars, blank lines in some places etc. :-)

Please, read carefully https://github.com/spring-projects/spring-integration/blob/master/CONTRIBUTING.adoc. We have there almost everything about that.

PR #162, opened 2016-07-06
Migrate leader election support from Spring Cloud

dsyer:


spring-integration-hazelcast/src/main/java/org/springframework/integration/hazelcast/leader/LeaderInitiator.java
```diff
@@ -0,0 +1,253 @@
...
+import org.springframework.integration.leader.DefaultCandidate;
+import org.springframework.integration.leader.event.DefaultLeaderEventPublisher;
+import org.springframework.integration.leader.event.LeaderEventPublisher;
+import org.springframework.util.Assert;
+
+import com.hazelcast.core.HazelcastInstance;
+import com.hazelcast.core.IMap;
+
+/**
+ * Bootstrap leadership {@link org.springframework.cloud.cluster.leader.Candidate candidates}
```
artembilan: Should be `org.springframework.integration.leader.Candidate` instead.

spring-integration-hazelcast/src/main/java/org/springframework/integration/hazelcast/leader/LeaderInitiator.java
```diff
@@ -0,0 +1,253 @@
...
+
+import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import java.util.concurrent.ThreadFactory;
+import java.util.concurrent.TimeUnit;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
```
artembilan: We don't use `slf4j` as a compiled logging system.
Will fix on merge, of course :smile:

spring-integration-hazelcast/src/main/java/org/springframework/integration/hazelcast/leader/LeaderInitiator.java
```diff
@@ -0,0 +1,253 @@
...
+
+/**
+ * Bootstrap leadership {@link org.springframework.cloud.cluster.leader.Candidate candidates}
+ * with Hazelcast. Upon construction, {@link #start} must be invoked to
+ * register the candidate for leadership election.
+ *
+ * @author Patrick Peralta
+ * @author Gary Russell
+ */
+public class LeaderInitiator implements Lifecycle, InitializingBean, DisposableBean, ApplicationEventPublisherAware {
```
artembilan: I think for consistency it must be `SmartLifecycle`. With that we eve let end-user control to autostartup or not.
Unless just unconditional:
```
public void afterPropertiesSet() throws Exception {
    start();
}
```
below.

spring-integration-hazelcast/src/main/java/org/springframework/integration/hazelcast/leader/LeaderInitiator.java
```diff
@@ -0,0 +1,253 @@
...
+		}
+	}
+
+	/**
+	 * Start the registration of the {@link #candidate} for leader election.
+	 */
+	@Override
+	public synchronized void start() {
+		if (!this.running) {
+			this.mapLocks = this.client.getMap("spring-cloud-leader");
```
artembilan: Yeah... We aren't in a `could` here any more. Although still Spring :smile: 

Anyway I think it would be better to expose property to let end-users to choose a target Hazelcast Map.
But, right, with something like `spring-leader` by default would enough.

artembilan: Or... Even better: just rely on the `beanName`.
I'm doing this variant. Let me know in case of any objections.

spring-integration-hazelcast/src/main/java/org/springframework/integration/hazelcast/leader/LeaderInitiator.java
```diff
@@ -0,0 +1,253 @@
...
+
+		@Override
+		public boolean isLeader() {
+			return LeaderInitiator.this.mapLocks != null && LeaderInitiator.this.mapLocks.isLocked(LeaderInitiator.this.candidate.getRole());
+		}
+
+		@Override
+		public void yield() {
+			if (LeaderInitiator.this.future != null) {
+				LeaderInitiator.this.future.cancel(true);
```
artembilan: Yeah... I see that the `yield` hasn't been fixed, as we discussion and had fixed in the `LockRegistryLeaderInitiator`: https://github.com/spring-projects/spring-integration/commit/1752204eb63c2edabde9c7fabb34cee76eb4535e
See commit comment since:
> Fix `yield()` logic based on the `Future.cancel(true)`.

I'll try to port my `yield()` here into Hazelcast module to prove the problem and than will try to fix it similar way via re-scheduling.

spring-integration-hazelcast/src/main/java/org/springframework/integration/hazelcast/leader/LeaderInitiator.java
artembilan: Well, I've changed my mind during development.
Since all we need a `lock` for the `role` we don't need any additional properties and even this `IMap` for locks.
There is just enough one distributed `ILock` for the `role` from candidate.

artembilan:
@dsyer , merged as https://github.com/spring-projects/spring-integration-extensions/commit/92a01dbe2969dda83c662ba48d90e92a7ab8bc76.
See commit comments and pay attention to code changes.

If you don't like something or I really have missed some case, you always can pull the master and PR us back again.

Thank you!
And let me guess: the next your contribution will be `HazelcastLockRegistry` :smile: 
And yes, here it would be based on `IMap` already.

Issue #164, opened 2016-09-09
ZipTransformer: "deleteFiles" throwing exception


muraliyaj:
Issue: 
`deleteFiles=true` causing below exception .
```
org.zeroturnaround.zip.ZipException: java.io.FileNotFoundException: <Abosulate Path of File with File name>The system cannot find the file specified)
```
Reason:
Inside `ZipTransformer`, `deleteFiles` is used to delete the original file before corresponding `FileSource` is used to compress the file.
That is causing `FileNotFoundException`.

artembilan:
Thank you for reporting this!

Really bug. Confirmed with simple `zipTransformer.setDeleteFiles(true);` modification to the `ZipTransformerTests.zipFile()` test-case.

Will be fixed soon.

muraliyaj:
Thank you Artem.

muraliyaj:
Also, This ZIP transformer may need 'error handling mechanism( like error output channel or hook points to advice chain)' to handle failed compression cases like disk full or issues with output channel.

artembilan:
The error handling isn't `Transfrormer` responsibility. The thrown Exception can be simply `try...catched` by the caller upstream or send to the `errorChannel` in case of async hand off.

In addition, with Java Config the `@Transformer` annotation can be supplied with `adviceChain` where one of that can be `ExpressionEvaluatingRequestHandlerAdvice`.
But, eh, I agree that we have missed `<request-hadnler-advice-chain>` there for the `<int-zip:zip-transformer>`.
Will fix here as well.

PR #165, opened 2016-09-09
GH-164: Fix `ZipTransformer` `deleteFile` mode

artembilan:
Fixes GH-164 (https://github.com/spring-projects/spring-integration-extensions/issues/164)

* Move `deleteFile` logic to the end of of `ZipTransformer#doZipTransform()`
* Add `deleteFile = true` for the `ZipTransformerTests`
* Add `<request-handler-advice-chain>` to the transformers XSD definitions
* Upgrade to Gradle 3.0
* Upgrade dependencies


spring-integration-zip/src/main/java/org/springframework/integration/zip/transformer/ZipTransformer.java
```diff
@@ -98,90 +98,103 @@ public void setUseFileAttributes(boolean useFileAttributes) {
...
+			zippedData = zippedFile;
+		}
+		else if (ZipResultType.BYTE_ARRAY.equals(this.zipResultType)) {
+			zippedData = zippedBytes;
+		}
+		else {
+			throw new IllegalStateException("Unsupported zipResultType " + this.zipResultType);
+		}
 
+		try {
```
garyrussell: What is the purpose of this `try {} finally {}` - how can building a message cause an exception and, even if it did, do we really want to delete the file after a failure?

artembilan: Ha! Sure thing :smile: 
Will move it before Message building.
Just thought that it would be logical to care about resource in the final end.
But, eh, that's true: isn't good idea to delete in case of error.
Doh...

garyrussell:
Merged as 9a13e41d9607517f978afdca4fa55f11a925ab4e

Issue #168, opened 2017-01-11
Do we have spring-integration-smb release version


agrawalchetan:
do we have release the spring integration smb version or is there any plan to release it in near future. with snapshot version, we are not able to download it and we are planning to have released version as we are moving our code to production environment.

artembilan:
Well, we can release that project as an early prototype in version like `0.5.0`.
But the problem that we will upgrade it to the latest dependencies versions.
Currently the project is based on the EOL Spring Integration `2.2.1`. 

So, the question to you: which Spring Integration version do you use in your project?
Would it be so difficult to you to upgrade to the latest `4.3.6`?
Since version `3.0.x` is out of support starting this year, I'm totally not thrilled to release SMB extension even with that version.

agrawalchetan:
Hi, we are using spring boot version 1.4.2  and spring-integration version 4.3.5. although we can move to 4.3.6 as well.

artembilan:
That's great!
So, we are upgrading `SI-SMB` project and releasing the next week!
That will be something like `0.5.0.RELEASE`.
Does that work for you?

agrawalchetan:
Thank you!!. please let me know once you will release the next version.

artembilan:
@agrawalchetan ,

please, find a PR #170 on the matter.
Let us know if you are fine with all those changes.

Thank you!

agrawalchetan:
thank you @artembilan, i am fine with the changes. please let me know release version artifact id once changes are done.

artembilan:
@agrawalchetan ,

We have just released `SI-SMB-0.5.0.RELEASE`: https://spring.io/blog/2017/01/18/spring-integration-extension-for-smb-0-5-0-is-available

Thank you for your request!

agk2:
@artembilan,
Thanks for releasing SI-SMB-0.5.0Release it really help full in our project.
Any plan to release SI-SMB-1.0.Release with SMB Outbound Channel Adapter?

artembilan:
@agk2 ,

We don't have requirements and demands on this project, so there is just no dedicated resources to go ahead. We all are busy with more important projects.
However contribution is always welcome and we will be more than happy to review, merge and eventually release improvements.

Thanks for the interest to the project any way!

PR #170, opened 2017-01-16
GH-168: Upgrade to SI-4.3.6 and others

artembilan:
Fixes GH-168 (https://github.com/spring-projects/spring-integration-extensions/issues/168)


garyrussell:
Travis was not enabled for this repo; I have enabled it now; if you re-push with a new hash, we should get one.

spring-integration-smb/build.gradle
```diff
@@ -1,43 +1,62 @@
...
+				mavenBom "io.spring.platform:platform-bom:${platformVersion}"
+			}
+		}
+	}
+}
+
+compileJava {
+	sourceCompatibility = 1.7
+	targetCompatibility = 1.7
 }
```
garyrussell: 1.7?

garyrussell:
Actually, I am not sure travis can work because each extension has its own build.

garyrussell:
Maybe put `.travis` at the top level and cd into the subproject?

Or add a top-level gradlw?

artembilan:
Sorry for delay and thank you for merging this one.
Well, since this project hasn't been released yet and it is an extension I think that shouldn't hurt to stick with Java 7 as minimum. Java is EOL already, even if SI 4.3.x is still with that 😄 

Regarding Travis, well, let me push some other PR an will see. At least I see Travis config in the Hazelcast directory. If it is there, that might be a case that Travis works.

I see that CI plan has to be fixed as well.

PR #171, opened 2017-01-16
SMB: Polishing XSD according latest SI

artembilan:


artembilan:
Hm. Really Travis doesn't work.

I don't have permissions to check the project config, but, yeah, we the closer look, it really doesn't feel like we can get it working. Even if we move `.travis.yml` to the root of the project, there is no hooks to understand to which sub-directory to jump to run real Gradle script for the current PR...

Issue #173, opened 2017-01-25
spring-integration-hazelcast LeaderInitiator shutting if error is catched during event propagation


philnate:
Hi,

I have the following scenario, where an exception is thrown during the LeaderEvent propagation. It seems like it's actually making it's way into the `org.springframework.integration.hazelcast.leader.LeaderInitiator` class, where it's not catched and even worse the error is swallowed and neither causes the app to die. This will leave you with an broken instance without further details to why it failed.

Not sure if I'm doing something wrong in my listener or if the LeaderInitiator should be more safeguarded around the LeaderEvent propagations.

artembilan:
I think it's really just an omission, but would be better to have some test-case to play with.

Thank you!

philnate:
Okay, let me put something together for you to play around.

philnate:
@artembilan please see https://github.com/philnate/issues/tree/spring-intex-hzlcst-173 for a minimal example of the described scenario.

artembilan:
OK. The problem is confirmed.
And that really doesn't look like event publishing error has to lead to revoke leadership.

Yes, I agree: the simple `warn` should be great compromise.

One of the workaround is an `executor` for the `SimpleApplicationEventMulticaster` to prevent exception bubbling from the event listener.
Another solution is a custom `LeaderEventPublisher`, probably just a simple extension of the `DefaultLeaderEventPublisher` to `try...catch` the target delegate invocation.

We would implement that in the `DefaultLeaderEventPublisher`, but that really from design perspective doesn't mean that we should allow to blow up granting by the event error.

The `try...catch` and warn.
That's all.

For reference: https://jira.spring.io/browse/INT-4220

PR #174, opened 2017-01-30
GH-173: Leaders: Warn event errors, not re-throw

artembilan:
Fixes GH-173 (https://github.com/spring-projects/spring-integration-extensions/issues/173)

Currently when an error is thrown from the event publishing the role granting is broken and we just go to the role revoking.

* Since it's just an event publishing it shouldn't effect the original leader election.
* `try...catch` event publishing in the `LeaderInitiator` and `logger.warn` an `Exception`

### DIFF
--- a/spring-integration-hazelcast/src/main/java/org/springframework/integration/hazelcast/leader/LeaderInitiator.java
+++ b/spring-integration-hazelcast/src/main/java/org/springframework/integration/hazelcast/leader/LeaderInitiator.java
@@ -23,6 +23,9 @@ import java.util.concurrent.Future;
 import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.TimeUnit;
 
+import org.apache.commons.logging.Log;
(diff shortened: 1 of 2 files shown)


artembilan:
@garyrussell , 

I'm pretty sure that we should port the solution to other `LeaderInitiator` implementations: https://jira.spring.io/browse/INT-4220

garyrussell:
I am not sure that this is the right answer on the grant side; if the application fails to process the leader onGranted event, shouldn't we yield leadership? On the revoked side it might be less of a problem because the app might still think it's leader, but I suppose there's not much more we can do than log the exception.

artembilan:
Why? It is just an event.
One of the listeners may process it properly, but another fail by some its own state.

For the strong leadership we require `Candidate`. The `LeaderEventPublisher` is optional for the `LeaderInitiator` anyway...

garyrussell:
Oh, yeah, I missed that - I forgot that leadership events are simply application event.

Issue #172, opened 2017-01-25
spring-integration-hazelcast requires JVM local Hazelcast instance


philnate:
Hi,

I tried to use the LeaderElection functionality for Hazelcast and stumpled upon the fact that `org.springframework.integration.hazelcast.HazelcastLocalInstanceRegistrar` is checking `com.hazelcast.core.Hazelcast` for all HazelcastInstances on that JVM. The thing is that this doesn't include HazelcastClients, but in our Scenario we have dedicated Hazelcast server to which we connect with HazelcastClients.

I lack the deeper knowledge of the spring-integration Project, as I'm just after the LeaderElection, but wouldn't it be sufficient to just have a HazelcastInstance autowired, which is then used to register the Listener on to?

For the moment I worked around the issue by tricking Spring into thinking that this Registrar is already present by adding a Dummy bean, which seems to do no harm for the LeaderElection scenario but likely for the other things.

```java
    @Bean( name = "org.springframework.integration.hazelcast.HazelcastLocalInstanceRegistrar" )
    public String HazelcastLocalInstanceRegistrar()
    {
        return "";
    }
```

artembilan:
OK. I think you have a point and we should make that local instances selection in the `HazelcastLocalInstanceRegistrar` as optional if an explicit instance is injected.
Right the `HazelcastLocalInstanceRegistrar` bean must be registered in that case.
So, I even guess that we should make an `id` for that bean much easier.

Sorry, for such an inconvenience.

PR #175, opened 2017-01-31
GH-172: Do not require local HZ instances

artembilan:
Fixes spring-projects/spring-integration-extensions#172

Since we can have an application based on the `HazelcastClient`, we don't need to require `Hazelcast.getAllHazelcastInstances()` be presented.

* Rework `HazelcastLocalInstanceRegistrar` to be able to accept external `HazelcastInstance` for `MultiMap` and `MembershipListener` registration
* If there is on local `Hazelcast.getAllHazelcastInstances()` just log a warn that we can't register `MembershipListener`
* Allow for `AbstractHazelcastMessageProducer` to accept events in the `CacheListeningPolicyType.SINGLE` mode when there is no local `Hazelcast.getAllHazelcastInstances()`


artembilan:
Merged as https://github.com/spring-projects/spring-integration-extensions/commit/5511987be258e6f7dbc3818237b059a832a92e7b

PR #180, opened 2017-07-14
Hz metadata store

viniciusccarvalho:
- Adds HazelcastMetadataStore

### DIFF
--- a/spring-integration-hazelcast/src/main/java/org/springframework/integration/hazelcast/metadata/HazelcastMetadataStore.java
+++ b/spring-integration-hazelcast/src/main/java/org/springframework/integration/hazelcast/metadata/HazelcastMetadataStore.java
@@ -1,0 +1,79 @@
+/*
+ * Copyright 2017 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+
+package org.springframework.integration.hazelcast.metadata;
+
+import com.hazelcast.core.HazelcastInstance;
+import com.hazelcast.core.IMap;
+
+import org.springframework.integration.metadata.ConcurrentMetadataStore;
+import org.springframework.util.Assert;
+
+/**
+ * @author Vinicius Carvalho
+ */
+public class HazelcastMetadataStore implements ConcurrentMetadataStore{
+
+	private static final String METADATA_STORE_MAP_NAME = "SPRING_INTEGRATION_METADATA_STORE";
+
+	private final IMap<String, String> map;
+
+	public HazelcastMetadataStore(HazelcastInstance hazelcastInstance){
+		Assert.notNull(hazelcastInstance, "Hazelcast instance can't be null");
+		this.map = hazelcastInstance.getMap(METADATA_STORE_MAP_NAME);
+	}
+
+	public HazelcastMetadataStore(IMap<String, String> map) {
+		Assert.notNull(map, "IMap reference can not be null");
+		this.map = map;
+	}
+
+	@Override
+	public String putIfAbsent(String key, String value) {
+		Assert.notNull(key, "'key' must not be null.");
+		Assert.notNull(value, "'value' must not be null.");
+		return this.map.putIfAbsent(key,value);
+	}
+
+	@Override
+	public boolean replace(String key, String oldValue, String newValue) {
+		Assert.notNull(key, "'key' must not be null.");
+		Assert.notNull(oldValue, "'oldValue' must not be null.");
+		Assert.notNull(newValue, "'newValue' must not be null.");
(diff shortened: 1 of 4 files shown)


spring-integration-hazelcast/src/test/java/org/springframework/integration/hazelcast/metadata/HazelcastMetadataStoreTests.java
```diff
@@ -0,0 +1,162 @@
...
+import org.junit.BeforeClass;
+import org.junit.Test;
+import org.mockito.ArgumentCaptor;
+
+import static org.mockito.Mockito.*;
+
+
+import org.springframework.integration.metadata.MetadataStoreListener;
+
+import static org.junit.Assert.*;
```
artembilan: Please, don't use asterisk. There is also some kind of trap when the class is imported from the unexpected package.

spring-integration-hazelcast/src/main/java/org/springframework/integration/hazelcast/metadata/HazelcastMetadataStore.java
```diff
@@ -0,0 +1,131 @@
...
+	}
+
+	@Override
+	public void removeListener(MetadataStoreListener callback) {
+		this.listeners.remove(callback);
+	}
+
+	@Override
+	public void afterPropertiesSet() throws Exception {
+		this.map.addEntryListener(new MapListener(),true);
```
artembilan: I would say we don't need this `afterPropertiesSet()`.
We have an `IMap` in the ctor and `listeners` is `final`, too.

Although, I'm pretty sure that with this `this.map.addEntryListener()` we go to the low level resource communication.
So, this is fine.

But... (!!!) At the same isn't `hazelcastInstance.getMap()` operation low-level resource access as well?
Maybe would be better to move it here in the `afterPropertiesSet()` as well?

I won't mind otherwise - just share your point of view, please!

spring-integration-hazelcast/src/test/java/org/springframework/integration/hazelcast/metadata/HazelcastMetadataStoreTests.java
```diff
@@ -0,0 +1,162 @@
...
+
+	@AfterClass
+	public static void destroy() throws Exception {
+		instance.shutdown();
+	}
+
+	@Before
+	public void setup() throws Exception{
+		this.metadataStore = new HazelcastMetadataStore(map);
+		metadataStore.afterPropertiesSet();
```
artembilan: `this.` must be used for properties access.
The Checkstyle concern.

spring-integration-hazelcast/src/main/java/org/springframework/integration/hazelcast/metadata/HazelcastMetadataStore.java
```diff
@@ -0,0 +1,131 @@
...
+	}
+
+	final class MapListener implements  EntryAddedListener<String,String>,
```
viniciusccarvalho: Don't think I fully understand why it has to be static. It will make access to the listeners via `this` not possible.

spring-integration-hazelcast/src/main/java/org/springframework/integration/hazelcast/metadata/HazelcastMetadataStore.java
viniciusccarvalho: Listener registration is at `afterPropertiesSet` because users could provide their own IMap via constructor and not set the listener. That was the only rationale.
I don't mind moving the `getMap` to afterPropertiesSet either.

spring-integration-hazelcast/src/main/java/org/springframework/integration/hazelcast/metadata/HazelcastMetadataStore.java
artembilan: Right, you also have to propagate `listeners` property via ctor.
The main reason to avoid extra hooks between outer/inner classes

garyrussell: >I can't recall now the source of this motivation

With a private inner class, if the CTOR is private, the compiler has to make a synthetic, package-protected, ctor that takes an extra dummy argument and delegates to the actual ctor. Making the ctor package protected avoids this extra call in the call stack.

We generally prefer static inner classes (unless there are a lot of dependencies on outer class fields) because they are more efficient. An extra argument (this) is added to every method call.

artembilan:
Regarding `GemfireMetadataStore` concern: https://jira.spring.io/browse/INT-4313

spring-integration-hazelcast/src/main/java/org/springframework/integration/hazelcast/metadata/HazelcastMetadataStore.java
```diff
@@ -0,0 +1,132 @@
...
+	public void removeListener(MetadataStoreListener callback) {
+		this.listeners.remove(callback);
+	}
+
+	@Override
+	public void afterPropertiesSet() throws Exception {
+		this.map.addEntryListener(new MapListener(), true);
+	}
+
+	private class MapListener implements EntryAddedListener<String, String>,
```
artembilan: LGTM
Will fix `static` concern on merge lately this evening...
👍

artembilan:
Merged as https://github.com/spring-projects/spring-integration-extensions/commit/89e6274cc577e4a8974e171a88d44906063111b0

@viniciusccarvalho , grate to have some contribution to this amazing Spring Integration extension!

Looking forward for more 😉

PR #182, opened 2017-10-22
Use FILENAME as zip file name if specified

jagedn:
If FILENAME is present at headers we must use this value as the name of the file without conversion


artembilan:
Thank you for the fix, @jagedn .

Looking forward for more!

PR #186, opened 2018-04-25
spring-integration-5.0.4 upgrade

prafsoni:


spring-integration-smb/src/test/java/org/springframework/integration/smb/config/SmbInboundChannelAdapterParserTests.java
```diff
@@ -75,14 +74,14 @@ public void testSmbInboundChannelAdapterComplete() throws Exception {
 		String remoteFileSeparator = (String) TestUtils.getPropertyValue(fisync, "remoteFileSeparator");
 		assertNotNull(remoteFileSeparator);
 		assertEquals("", remoteFileSeparator);
-		SmbSimplePatternFileListFilter filter = (SmbSimplePatternFileListFilter) TestUtils.getPropertyValue(fisync, "filter");
-		assertNotNull(filter);
+		//SmbSimplePatternFileListFilter filter = (SmbSimplePatternFileListFilter) TestUtils.getPropertyValue(fisync, "filter");
```
prafsoni: @artembilan I assume this behavior change is due to Spring integration changes. If I am wrong kindly, guide me how can revert this.

spring-integration-smb/build.gradle
```diff
@@ -171,8 +171,8 @@ task schemaZip(type: Zip) {
 	description = "Builds -${classifier} archive containing all " +
 		"XSDs for deployment at static.springframework.org/schema."
 
-	def Properties schemas = new Properties();
-	def shortName = idPrefix.replaceFirst("${idPrefix}-", '')
+    Properties schemas = new Properties()
+    def shortName = idPrefix.replaceFirst("${idPrefix}-", '')
```
artembilan: I think this is wrong indent. We prefer tabs

spring-integration-smb/src/main/java/org/springframework/integration/smb/filters/SmbPersistentAcceptOnceFileListFilter.java
```diff
@@ -0,0 +1,45 @@
+/*
+ * Copyright 2012-2017 the original author or authors.
```
artembilan: This is a new class in this year, so must be only `2018`.

All other affected classes must be changes to reflect the current year in their second number.

Plus all the affected classes should have your name now! :smile:

spring-integration-smb/src/main/java/org/springframework/integration/smb/filters/SmbPersistentAcceptOnceFileListFilter.java
```diff
@@ -0,0 +1,45 @@
...
+
+import jcifs.smb.SmbFile;
+import org.springframework.integration.file.filters.AbstractPersistentAcceptOnceFileListFilter;
+import org.springframework.integration.metadata.ConcurrentMetadataStore;
+
+/**
+ * Implementation of {@link AbstractPersistentAcceptOnceFileListFilter} for SMB.
+ *
+ * @author Prafull Kumar Soni
+ * @since 0.6.0
```
artembilan: I think we don't need this `@since`.
When we merge your fix and will do some other polishing, we will release `1.0.0.RELEASE` to shot the noise in the public.

spring-integration-smb/src/main/java/org/springframework/integration/smb/filters/SmbRegexPatternFileListFilter.java
```diff
@@ -48,4 +51,19 @@ protected String getFilename(SmbFile file) {
 		return (file != null ? file.getName() : null);
 	}
 
+	/**
+	 * Indicates whether the file is a directory or not.
+	 * @param file SMB file object
+	 * @return true if it's a directory.
+	 * @see AbstractDirectoryAwareFileListFilter#isDirectory(java.lang.Object)
```
artembilan: I think we don't need JavaDocs on this method at all. It is inherited.

More over, please, don't use imports if they are mentioned only in JavaDocs.

spring-integration-smb/src/main/java/org/springframework/integration/smb/filters/SmbSimplePatternFileListFilter.java
```diff
@@ -43,4 +47,19 @@ protected String getFilename(SmbFile file) {
...
+	 * @return true if it's a directory.
+	 * @see AbstractDirectoryAwareFileListFilter#isDirectory(java.lang.Object)
+	 */
+	@Override
+	protected boolean isDirectory(SmbFile file) {
+		try {
+			return file.isDirectory();
+		}
+		catch (SmbException e) {
+			throw new UncheckedIOException(e);
```
artembilan: That's cool, of course, long waiting runtime exception for IO :smiley: 

But I wonder if there is a new version of SMB with better Java 8 adoption that we can just upgrade and go.

spring-integration-smb/src/main/java/org/springframework/integration/smb/filters/SmbSystemMarkerFilePresentFileListFilter.java
```diff
@@ -0,0 +1,54 @@
...
+import org.springframework.integration.file.filters.FileListFilter;
+
+import java.util.Map;
+import java.util.function.Function;
+
+/**
+ * Implementation of {@link AbstractMarkerFilePresentFileListFilter} for SMB.
+ *
+ * @since 0.6.0
+ * @author Prafull Kumar Soni
```
artembilan: `@since` goes after an `@author`.
Anyway we discussed do not `@since` in the current version.

spring-integration-smb/src/test/java/org/springframework/integration/smb/config/SmbInboundChannelAdapterParserTests.java
artembilan: Yeah... That's right: starting with some version (https://jira.spring.io/browse/INT-4115), we populate there a `CompositeFileListFilter`. 
You should do something like this here:
```
FileListFilter<?> filter = TestUtils.getPropertyValue(fisync, "filter", FileListFilter.class);
assertNotNull(filter);
assertThat(filter, instanceOf(CompositeFileListFilter.class));
Set<?> fileFilters = TestUtils.getPropertyValue(filter, "fileFilters", Set.class);

Iterator<?> filtersIterator = fileFilters.iterator();
assertThat(filtersIterator.next(), instanceOf(SmbSimplePatternFileListFilter.class));
assertThat(filtersIterator.next(), instanceOf(SmbPersistentAcceptOnceFileListFilter.class));
```

spring-integration-smb/src/main/java/org/springframework/integration/smb/filters/SmbSimplePatternFileListFilter.java
prafsoni: Yeah, I am not sure about that, there are other libs out there but I highly doubt they are Java 8 

<https://github.com/hierynomus/smbj> is Java 7 and supports SMB2 and SMB3 vs current jCIFS supports CIFS/SMB. Apache Commons VFS also have CIFS (sandbox).

Let me know your thoughts. I think it will not be too hard to swap out the libs.

spring-integration-smb/src/test/java/org/springframework/integration/smb/config/SmbInboundChannelAdapterParserTests.java
prafsoni: Thanks, I will get this updated.

prafsoni:
Cool, I will get all these changes done tonight.

spring-integration-smb/src/main/java/org/springframework/integration/smb/filters/SmbSimplePatternFileListFilter.java
artembilan: Well, let's fix this issue and maybe release with what we have so far.
And the choice to move to some more modern library let's do afterwards!

prafsoni:
@artembilan Done

PR #189, opened 2018-05-03
Disallow traversal entity in zip

artembilan:
When the file name holds path traversal file names it gets
concatenated to the target extraction directory,
the final path ends up outside of the target folder.


spring-integration-zip/src/main/java/org/springframework/integration/zip/transformer/UnZipTransformer.java
```diff
@@ -131,6 +133,16 @@ public void process(InputStream zipEntryInputStream, ZipEntry zipEntry) throws I
...
 							final File destinationFile = new File(tempDir, zipEntryName);
 
+							/* If we see the relative traversal string of ".." we need to make sure
+							 * that the outputdir + name doesn't leave the outputdir.
+							 */
+							if (zipEntryName.contains("..") &&
+									!destinationFile.getCanonicalPath().startsWith(workDirectory.getCanonicalPath())) {
+								throw new ZipException("The file " + zipEntryName +
+										" is trying to leave the target output directory of " + workDirectory
+										+ ". Ignoring this file.");
```
garyrussell: Here, we're not just ignoring this file, we're rejecting the entire zip file (which is probably the best thing to do). Just needs a fix to the exception message, to remove the last sentence.

artembilan: Yes, please.

artembilan:
Merged as https://github.com/spring-projects/spring-integration-extensions/commit/a5573eb232ff85199ff9bb28993df715d9a19a25

Issue #206, opened 2019-04-06
Update version of dependant library jCIFS to handle SMB2/SMB3


GregBragg:
Currently the version of jCIFS used only supports SMB1, need to update the dependant lib to the latest version that handles SMB2/SMB3.

artembilan:
Here is the release BTW: https://spring.io/blog/2019/04/10/spring-integration-smb-1-1-ga-available

GregBragg:
Thanks you very much Artem!

Greg.

PR #207, opened 2019-04-08
Updated jCIFS library to the latest version to support SMB2/SMB3

GregBragg:
Fixes https://github.com/spring-projects/spring-integration-extensions/issues/206

Updated dependent jCIFS library to support SMB2 and SMB3 protocols since SMB1 is no longer supported by the latest versions of Windows servers as it was deemed a security risk by Microsoft. SMB1 has been deprecated.

### DIFF
--- a/spring-integration-smb/build.gradle
+++ b/spring-integration-smb/build.gradle
@@ -56,7 +56,7 @@ compileTestJava {
 ext {
 	idPrefix = 'smb'
 
-	jcifsVersion = '1.3.18.3'
+	jcifsVersion = '2.1.7'
 	log4jVersion = '2.11.0'
 	springIntegrationVersion = '5.0.13.BUILD-SNAPSHOT'
 
--- a/spring-integration-smb/src/main/java/org/springframework/integration/smb/session/SmbSession.java
+++ b/spring-integration-smb/src/main/java/org/springframework/integration/smb/session/SmbSession.java
@@ -63,10 +63,6 @@ public class SmbSession implements Session<SmbFile> {
 
 	private static final String SMB_FILE_SEPARATOR = "/";
 
-	static {
-		configureJcifs();
-	}
-
 	private final SmbShare smbShare;
 
 	/**
@@ -458,7 +454,7 @@ public class SmbSession implements Session<SmbFile> {
 			return this.smbShare;
 		}
 
-		SmbFile smbFile = new SmbFile(this.smbShare, cleanedPath, SmbFile.FILE_SHARE_READ);
+		SmbFile smbFile = new SmbFile(this.smbShare, cleanedPath);
 
 		boolean appendFileSeparator = !cleanedPath.endsWith(SMB_FILE_SEPARATOR);
 		if (appendFileSeparator) {
@@ -496,32 +492,6 @@ public class SmbSession implements Session<SmbFile> {
 	 */
 	public SmbFile createSmbDirectoryObject(String _path) throws IOException {
 		return createSmbFileObject(_path, true);
-	}
-
-	/**
-	 * Static configuration of the JCIFS library.
-	 * The log level of this class is mapped to a suitable <code>jcifs.util.loglevel</code>
-	 */
-	static void configureJcifs() {
-		// TODO jcifs.Config.setProperty("jcifs.smb.client.useExtendedSecurity", "false");
-		// TODO jcifs.Config.setProperty("jcifs.smb.client.disablePlainTextPasswords", "false");
-
-		// set JCIFS SMB client library' log level unless already configured by system property
-		final String sysPropLogLevel = "jcifs.util.loglevel";
-
-		if (jcifs.Config.getProperty(sysPropLogLevel) == null) {
-			// set log level according to this class' logger's log level.
-			Log log = LogFactory.getLog(SmbSession.class);
-			if (log.isTraceEnabled()) {
-				jcifs.Config.setProperty(sysPropLogLevel, "N");
-			}
-			else if (log.isDebugEnabled()) {
-				jcifs.Config.setProperty(sysPropLogLevel, "3");
-			}
-			else {
-				jcifs.Config.setProperty(sysPropLogLevel, "1");
-			}
-		}
 	}
 
 	@Override
--- a/spring-integration-smb/src/test/java/org/springframework/integration/smb/session/SmbSessionTests.java
+++ b/spring-integration-smb/src/test/java/org/springframework/integration/smb/session/SmbSessionTests.java
@@ -27,81 +27,110 @@ import jcifs.smb.SmbFile;
 /**
  *
  * @author Gunnar Hillert
+ * @author Gregory Bragg - 2019/04/09 - All tests rewritten during the upgrade to the latest version of jCIFS
  *
  */
 public class SmbSessionTests {
 
 	@Test
 	public void testCreateSmbFileObjectWithBackSlash1() throws IOException {
-
 		System.setProperty("file.separator", "\\");
-		SmbShare smbShare = new SmbShare("smb://myshare/shared/");
+		SmbConfig config = new SmbConfig();
+		config.setHost("myshare");
+		config.setPort(445);
+		config.setShareAndDir("shared/");
+		SmbShare smbShare = new SmbShare(config);
 		SmbSession smbSession = new SmbSession(smbShare);
+
 		SmbFile smbFile = smbSession.createSmbFileObject("smb://myshare\\blubba\\");
 		assertEquals("smb://myshare/blubba/", smbFile.getPath());
+		smbSession.close();
 	}
 
 	@Test
 	public void testCreateSmbFileObjectWithBackSlash2() throws IOException {
-
 		System.setProperty("file.separator", "\\");
-		SmbShare smbShare = new SmbShare("smb://myshare\\shared\\");
+		SmbConfig config = new SmbConfig();
+		config.setHost("myshare");
+		config.setPort(445);
+		config.setShareAndDir("shared\\");
+		SmbShare smbShare = new SmbShare(config);
 		SmbSession smbSession = new SmbSession(smbShare);
+
 		SmbFile smbFile = smbSession.createSmbFileObject("smb://myshare\\blubba\\");
 		assertEquals("smb://myshare/blubba/", smbFile.getPath());
+		smbSession.close();
 	}
 
 	@Test
 	public void testCreateSmbFileObjectWithBackSlash3() throws IOException {
-
 		System.setProperty("file.separator", "\\");
-		SmbShare smbShare = new SmbShare("smb://myshare\\shared\\");
+		SmbConfig config = new SmbConfig();
+		config.setHost("myshare");
+		config.setPort(445);
+		config.setShareAndDir("shared\\");
+		SmbShare smbShare = new SmbShare(config);
 		SmbSession smbSession = new SmbSession(smbShare);
+
 		SmbFile smbFile = smbSession.createSmbFileObject("..\\another");
-		assertEquals("smb://myshare/another", smbFile.getPath());
+		assertEquals("smb://myshare:445/another", smbFile.getPath());
+		smbSession.close();
 	}
 
 	@Test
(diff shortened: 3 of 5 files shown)


spring-integration-smb/src/main/java/org/springframework/integration/smb/session/SmbSession.java
```diff
@@ -458,7 +454,7 @@ private SmbFile createSmbFileObject(String path, Boolean isDirectory) throws IOE
 			return this.smbShare;
 		}
 
-		SmbFile smbFile = new SmbFile(this.smbShare, cleanedPath, SmbFile.FILE_SHARE_READ);
+		SmbFile smbFile = new SmbFile(this.smbShare, cleanedPath);
```
artembilan: Why have you dropped that `SmbFile.FILE_SHARE_READ` option?
Isn't there some alternative any more?

spring-integration-smb/src/main/java/org/springframework/integration/smb/session/SmbSession.java
```diff
@@ -498,32 +494,6 @@ public SmbFile createSmbDirectoryObject(String _path) throws IOException {
 		return createSmbFileObject(_path, true);
 	}
 
-	/**
-	 * Static configuration of the JCIFS library.
-	 * The log level of this class is mapped to a suitable <code>jcifs.util.loglevel</code>
-	 */
-	static void configureJcifs() {
```
artembilan: So, don't we need to reconfigure logging for `jcifs` any more?
What drove you to drop this off?

spring-integration-smb/src/test/java/org/springframework/integration/smb/session/SmbSessionTests.java
```diff
@@ -38,7 +38,7 @@ public void testCreateSmbFileObjectWithBackSlash1() throws IOException {
 		SmbShare smbShare = new SmbShare("smb://myshare/shared/");
 		SmbSession smbSession = new SmbSession(smbShare);
 		SmbFile smbFile = smbSession.createSmbFileObject("smb://myshare\\blubba\\");
-		assertEquals("smb://myshare/blubba/", smbFile.getCanonicalPath());
+		assertEquals("smb://myshare/shared/smb://myshare/blubba/", smbFile.getCanonicalPath());
```
artembilan: I think this is something wrong. It definitely must not so creepy path...

That `/blubba` should go to the sub-path of the `smb://myshare/shared/` root.
Isn't it?

spring-integration-smb/src/test/java/org/springframework/integration/smb/session/SmbSessionTests.java
```diff
@@ -58,7 +58,7 @@ public void testCreateSmbFileObjectWithBackSlash3() throws IOException {
 		SmbShare smbShare = new SmbShare("smb://myshare\\shared\\");
 		SmbSession smbSession = new SmbSession(smbShare);
 		SmbFile smbFile = smbSession.createSmbFileObject("..\\another");
-		assertEquals("smb://myshare/another/", smbFile.getCanonicalPath());
+		assertEquals("smb://myshare/shared/../another", smbFile.getCanonicalPath());
```
artembilan: This is something wrong, too.
If we request to create a file in the root of the `smb://myshare`, then we need to pursue that anyway.
What happened with relative path resolution in this new client?

spring-integration-smb/src/main/java/org/springframework/integration/smb/session/SmbSession.java
GregBragg: Well they are using SLFJ now and not their embedded, homegrown logger that wasn't really compliant with Log4J and Commons Log. We can configure the logging level easily using config files now in JCIFS.

artembilan: Good. Would be great to reflect that in the commit message.
Thanks

spring-integration-smb/src/test/java/org/springframework/integration/smb/session/SmbSessionTests.java
GregBragg: The getCanonicalPath() method returns a completely different value now, verified by stepping it through the debugger. I'm not sure what was being asserted here for the paths so maybe the method call needs to change instead of the data being checked?

spring-integration-smb/src/test/java/org/springframework/integration/smb/session/SmbSessionTests.java
GregBragg: Same comment as above.

spring-integration-smb/src/main/java/org/springframework/integration/smb/session/SmbSession.java
GregBragg: Sorry should have stated that... Thanks!

spring-integration-smb/src/test/java/org/springframework/integration/smb/session/SmbSessionTests.java
GregBragg: It looks like the test needs a few changes, getPath() method instead of getCanonicalPath() and some of the data needs to change too such as removal of a forward slash at the end of the string for 3 tests.
Let me know if you want this updated. Thanks!

spring-integration-smb/src/test/java/org/springframework/integration/smb/session/SmbSessionTests.java
artembilan: Yes, please, address that concern for better reflection of reality.

Thanks

GregBragg: Ok I've update the SmbSessionTests to use getPath() instead of getCanonicalPath(). It looks much better and appears to assert the data that this test was originally checking for.

spring-integration-smb/src/main/java/org/springframework/integration/smb/session/SmbSession.java
artembilan: Any insights here, please?

GregBragg: The constructor for SmbFile( SmbFile context, String name, int shareAccess ) has been removed from the latest version of jCIFS, and the closest one now available is public SmbFile ( SmbResource context, String name ).  SmbFile now implements SmbResource. SmbResource is also new.

artembilan: That's OK with me, but how about that `shareAccess`? Isn't there a way to configure it any more?
How can we be sure that an appropriate access is granted?

GregBragg: From what I can tell by reading through the jCIFS code that the share attribute was used in a constructor to the old SmbComNTCreateAndX class that is now in a separate internal\smb1 package off the root jcifs package. I think that is why this old constructor was removed. It might be redundant now. I'm still looking to see if there is a comparable way to set the share to "file share read".

GregBragg: Ok pouring through the latest code for jCIFS appears to show that this was an old concept for SMB1, hence the reason it was removed. The new constructor without the SmbFile.FILE_SHARE_READ flag initializes a SMB2 file with flags SmbConstants.FILE_SHARE_READ | SmbConstants.FILE_SHARE_WRITE. It looks like this old feature from SMB1 has been disabled.

The old SmbFile has been moved into a new jcifs.smb1.smb1 package along with all the old code for SMB1. Since we are upgrading this extension to use the latest SMB protocols then I think it is no longer necessary to set the share access permissions. The latest version doesn't appear to allow this. Also all the JUnit tests in jCIFS do not set this share flag anymore, nor is there an example I can use to set this flag for SmbConstants.FILE_SHARE_READ only.

artembilan: OK. Fair enough.

I really can just live with whatever is there by default.

GregBragg:
I've updated the readme with some verbiage from https://spring.io/blog/2017/01/18/spring-integration-extension-for-smb-0-5-0-is-available so I hope you don't mind. It helped me out when I first started using this extension. Thanks!

GregBragg:
Yes I saw that too and I was looking into it... It seems the authors of JCIF deprecated a whole class NtlmPasswordAuthentication, without any reason or suggestions of what else to use. I'll see what I can do.

spring-integration-smb/build.gradle
```diff
@@ -56,7 +56,7 @@ compileTestJava {
 ext {
 	idPrefix = 'smb'
 
-	jcifsVersion = '1.3.18.3'
+	jcifsVersion = '2.1.7'
 	log4jVersion = '2.11.0'
 	springIntegrationVersion = '5.0.13.BUILD-SNAPSHOT'
```
garyrussell: So you need this version of SMB to work with this old Spring Integration version?

If not, we should bump this to 5.1.4.RELEASE.

artembilan: OK with me, but let's address that in the separate issue!
Seems for me there is some incompatibility.

garyrussell: Ok; makes sense; but we **will** need to bump `gradle.properties` to 1.1.0.BUILD-SNAPSHOT for this dependency change.

artembilan: Yep! And I guess we even can release this extension afterward to meet community requirements and make their lives easier with SMB 2/3 support 😄

GregBragg:
Question: Why do we have this constructor in the code when the only class that is using it is SmbSessionTests? I think we can also deprecate it safely and I'll rewrite the other constructor to authenticate without using the other deprecated class in jCIFS.

	public SmbShare(String url) throws IOException {
		super(StringUtils.cleanPath(url));
	}

artembilan:
Right, if underlying library doesn't provide such a feature any more, let's deprecate it from our side as well!

Thanks for digging all the out!

GregBragg:
Ok well do you want me to change the test class also? Or can it be ignored?

artembilan:
Please, fix it. No reason to remove anything yet.

Thanks

GregBragg:
The deprecated constructor issue in jCIFS has been addressed. I tested the snapshot build that I created in my own batch integration that this PR is now depending on with full debugging turned on and it is working perfectly for an SMB2 connection to a file server share.

spring-integration-smb/build.gradle
GregBragg: We are currently using the 5.0.3.RELEASE right now. Is there much difference between this version and the one you are proposing?

garyrussell: >We are currently using the 5.0.3.RELEASE 

@GregBragg you should really try to keep more up-to-date with releases, at least within a minor release (5.0.x). 5.0.3 is over a year old 5.0.13 represents 10 bug-fix releases since then (mostly minor, but some are important).

While I would prefer bumping this to 5.1.x, I don't have a major objection to releasing this against 5.0.13 and quickly following up with 1.2.0.RELEASE against 5.1.4 and 1.3.0.BUILD-SNAPSHOT against 5.2.x (which will get its first milestone soon).

(Thanks for all your work on this, BTW).

GregBragg:
Thanks! Much appreciated...

artembilan:
Merged as https://github.com/spring-projects/spring-integration-extensions/commit/376a34c03128c013ae063a0b1a3eae51fa3e176d after some minor polishing and upgrades.

@GregBragg ,

Thank you very much for the contribution; looking forward for more!

GregBragg:
Your very welcome! When can I expect the new 1.1.0.RELEASE build created and pushed to the Maven Repo?

artembilan:
Let's hope today!

We have some stuff to do in other project and release that as well, then we can come back to this one.

If not today, then definitely on Friday, since we have more to release tomorrow.

Thanks for understanding!

Issue #225, opened 2020-06-20
spring-integration-smb memory leak in SmbShare


beardy247:
Hi,

I've been experiencing a memory leak generated by SmbSession. I wasn't using a CachingSessionFactory so each poll was causing a leak due to the open file blocking the thread from finalizing. I've tracked it down to this method on SmbShare:

```
void doClose() {
    this.open.set(false);
}
```

Looks like the comment is out of date, jcifs.smb.SmbFile close() method is public. Calling super.close() resolves the memory leak. 

Thanks,
Adam

artembilan:
Thank you for reporting this observation! We would appreciate if you raise a pull request with respective fix.

PR #227, opened 2020-06-22
Fix for memory leak issue #225

beardy247:
If SmbShare instantiates a BaseContext during construction then the
context will be closed on calling close().

Without this there's a ThreadLocal memory leak.

### DIFF
--- a/spring-integration-smb/src/main/java/org/springframework/integration/smb/session/SmbSession.java
+++ b/spring-integration-smb/src/main/java/org/springframework/integration/smb/session/SmbSession.java
@@ -406,7 +406,7 @@ public class SmbSession implements Session<SmbFile> {
 
 	@Override
 	public void close() {
-		this.smbShare.doClose();
+		this.smbShare.close();
 	}
 
 	/**
--- a/spring-integration-smb/src/main/java/org/springframework/integration/smb/session/SmbShare.java
+++ b/spring-integration-smb/src/main/java/org/springframework/integration/smb/session/SmbShare.java
@@ -28,6 +28,7 @@ import org.springframework.util.Assert;
 import org.springframework.util.StringUtils;
 
 import jcifs.CIFSContext;
+import jcifs.CIFSException;
 import jcifs.config.PropertyConfiguration;
 import jcifs.context.BaseContext;
 import jcifs.context.SingletonContext;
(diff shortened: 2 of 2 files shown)


beardy247:
Hi @artembilan,
I've deleted the previous branch and fixed the issue with line endings in git/eclipse that caused the issue with the old pull request.
I've checked this thoroughly with JProfiler and I'm now satisfied that the leak when polling is gone.
Thanks,
Adam

artembilan:
Thank you for contribution ; looking forward for more!

beardy247:
When's the next release scheduled?

GregBragg:
@artembilan we might want to upgrade the base jCIFS library to the latest release before we create a new build as we are 7 minor,minor versions behind.

@beardy247 maybe you might want to reviewed these changes to the base library:  https://github.com/codelibs/jcifs/compare/jcifs-2.1.11...jcifs-2.1.18 and retest with JProfiler?

I see a few fixes for "Fix session/connection leak on authentication errors (#191)" and "Lazily remove connections from pool to avoid deadlocks (#188)" that might resolve further memory/thread issues.

beardy247:
@GregBragg I'll take a look, I've got a few days left on my JProfiler trial. I did spot the leak on authentication errors, but forgot to pursue it as I was focusing on the main leak as it was a major blocker to the progress of our project.

GregBragg:
Awesome! Thanks for this, really appreciate the extra effort!

beardy247:
@GregBragg I've done some profiling and I was all but convinced there was still a leak in the jCIFS library on authentication errors but on closer inspection I think the library is OK. To make it easier to spot leaks in JProfiler I'd taken out the CachingSessionFactory from my code so that each poll created a new session (with invalid credentials) with a polling period of 10 seconds, I think the leak stems from the default timeout in SmbTransportImpl being 30 seconds, so it's creating sessions faster than it's tidying up after itself. The answer may be just to call getContext().close() in the catch block of SmbShare.init(). What do you think?

beardy247:
@GregBragg Having looked again at this this morning, I think the timeout is a red herring. I've submitted a new pull request with a fix that protects spring integration from the leak.

beardy247:
I've reviewed the other changes to the base library and they all look OK.

GregBragg:
@beardy247 Thanks for your contribution! It looks great.

PR #228, opened 2020-06-24
Fix memory leak on authentication error

beardy247:
I believe the root cause of this is in the jCIFS library but this mitigates the
problem.

### DIFF
--- a/spring-integration-smb/build.gradle
+++ b/spring-integration-smb/build.gradle
@@ -42,7 +42,7 @@ compileTestJava {
 ext {
 	idPrefix = 'smb'
 
-	jcifsVersion = '2.1.18'
+	jcifsVersion = '2.1.19'
 	log4jVersion = '2.12.1'
 	springIntegrationVersion = '5.2.1.RELEASE'
 
--- a/spring-integration-smb/build.gradle
+++ b/spring-integration-smb/build.gradle
@@ -42,7 +42,7 @@ compileTestJava {
 ext {
 	idPrefix = 'smb'
 
-	jcifsVersion = '2.1.11'
+	jcifsVersion = '2.1.18'
 	log4jVersion = '2.12.1'
 	springIntegrationVersion = '5.2.1.RELEASE'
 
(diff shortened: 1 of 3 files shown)


beardy247:
@GregBragg @artembilan

beardy247:
Note: I didn't include the upgrade to jcifs-2.1.18 in this commit but that is the version that I'm using.

GregBragg:
@beardy247 Ok thanks, looks good! You can commit the changes to upgrade the jCIFS library to the latest release which as of yesterday is jcifs-2.1.19 in the Gradle build.gradle file if you want. I'm good with the changes.

GregBragg:
@beardy247  Actually scratch my last comment, I only see  jcifs-2.1.18 in the Maven Repository. Looks like they are only preparing to build the next release as it is not ready yet.

beardy247:
@GregBragg I've upgraded to 2.1.18 as per your message about 2.1.19 not being in the maven repo yet. Not sure why the message isn't in the thread above.

GregBragg:
@beardy247 yeah sorry, I deleted the message as I couldn't find the library at first in Maven, then I found it.

I reviewed the change that was made from 2.1.18 to 2.1.19 and it's not significant enough for me to ask you to retest with 2.1.19, however I'll leave that up to you if you want to make the extra effort.

The only change was this:
https://github.com/codelibs/jcifs/commit/be2dc950404d14af828b14036211d9447146e324
so not sure if that makes any difference in what you were seeing with the memory leaks.

GregBragg:
I don't think it would hurt Artem. The way that the URLStreamHandler was instantiated before in version 2.1.18 was suspect anyways, I like the fix they did from 2.1.18 to 2.1.19, so I think we should go with the latest build which only came out yesterday.

beardy247:
I've tested with 2.1.19 and the change is inconsequential to our usage.

GregBragg:
@beardy247 Thanks a bunch!

beardy247:
I've got some additional stuff to contribute in the future, I've written a DSL wrapper for the  SmbInboundFileSynchronizer and a SmbLastModifiedFileListFilter. I can't submit them yet though as I haven't had change to write unit tests for them. When I find some time I'll get the tests written and a pull request sorted.

artembilan:
@beardy247 , @GregBragg ,

`spring-integration-smb-1.2.1.RELEASE` is on its way to Maven Central.

Issue #235, opened 2020-10-14
Memory leak with org.springframework.integration.hazelcast.leader.LeaderInitiator


mleguevel:
Hello,

I think that the hazelcast [LeaderInitiator](https://github.com/spring-projects/spring-integration-extensions/blob/462a469d4c755021ba7a1730ddc86910692093e1/spring-integration-hazelcast/src/main/java/org/springframework/integration/hazelcast/leader/LeaderInitiator.java#L276) causes a memory leak. The reason for this is that it's calling `tryLock` twice before calling `unlock` and continuing with `tryLock` / `unlock`.

When doing this, an internal Map in hazelcast keeps growing while the LeaderInitiator keeps calling `tryLock` and `unlock`. This Map can be found either in [com.hazelcast.cp.internal.datastructures.lock.RaftLock](https://github.com/hazelcast/hazelcast/blob/ec3629d281c98c9671c70ec5f039da58312baa9c/hazelcast/src/main/java/com/hazelcast/cp/internal/datastructures/lock/RaftLock.java#L69) for the version 3.12.9 of hazelcast or in [com.hazelcast.cp.internal.datastructures.lock.Lock](https://github.com/hazelcast/hazelcast/blob/2f2e67ab1456b69ef1b6e59cfb7e71b070f7e8fe/hazelcast/src/main/java/com/hazelcast/cp/internal/datastructures/lock/Lock.java#L70) for the version 4.0.3. I've reproduced the problem with both versions.

I first created [an issue in hazelcast repository](https://github.com/hazelcast/hazelcast/issues/17709) and [their conclusion](https://github.com/hazelcast/hazelcast/issues/17709#issuecomment-707792757) seems to be that it works as designed.

Can this be considered a bug in LeaderInitiator then?

Thank you.

mleguevel:
Hi @artembilan 

I created a PR: https://github.com/spring-projects/spring-integration-extensions/pull/236
I'm not entirely sure that it's totally equivalent to the previous algorithm though. But the LeaderInitiatorTest tests pass.
I removed the two additional hazelcast instances from the test class because I assumed that they were added to circumvent the bug that prevented hazelcast from starting in unsafe mode in the 3.12.X versions. Tell me if it's not the reason.

PR #236, opened 2020-10-15
Fix a memory leak in hazelcast LeaderInitiator

mleguevel:
According to the hazelcast team: "The logic assumes that locks are generally acquired & released in a fairly short time or hold a very long time without unlocking. But in this case, is a bit different, it holds the lock for a long time but also does lock/unlock very frequently". The previous implementation used the described logic, first acquiring a lock and then doing frequent tryLock/unlock. Doing this leads to com.hazelcast.cp.internal.datastructures.lock.Lock#ownerInvocationRefUids to grow without ever being cleaned thus leading to an OutOfMemory error eventually.

### DIFF
--- a/spring-integration-hazelcast/src/main/java/org/springframework/integration/hazelcast/leader/LeaderInitiator.java
+++ b/spring-integration-hazelcast/src/main/java/org/springframework/integration/hazelcast/leader/LeaderInitiator.java
@@ -272,33 +272,19 @@ public class LeaderInitiator implements SmartLifecycle, DisposableBean, Applicat
 			try {
 				while (isRunning()) {
 					try {
-						// We always try to acquire the lock, in case it expired
-						boolean acquired =
-								LeaderInitiator.this.lock.tryLock(LeaderInitiator.this.heartBeatMillis,
-										TimeUnit.MILLISECONDS);
-						if (!this.locked) {
-							if (acquired) {
-								// Success: we are now leader
-								this.locked = true;
-								handleGranted();
-							}
-						}
-						else if (acquired) {
-							// If we were able to acquire it but we were already locked we
-							// should release it
-							LeaderInitiator.this.lock.unlock();
-							if (isRunning()) {
-								// Give it a chance to expire.
-								Thread.sleep(LeaderInitiator.this.heartBeatMillis);
-							}
-						}
-						else {
-							this.locked = false;
-							// We were not able to acquire it, therefore not leading any more
-							handleRevoked();
-							if (isRunning()) {
-								// Try again quickly in case the lock holder dropped it
-								Thread.sleep(LeaderInitiator.this.busyWaitMillis);
+						if (LeaderInitiator.this.lock.isLocked()) {
+							Thread.sleep(LeaderInitiator.this.heartBeatMillis);
+						} else {
+							// We try to acquire the lock
+							boolean acquired =
+									LeaderInitiator.this.lock.tryLock(LeaderInitiator.this.heartBeatMillis,
+											TimeUnit.MILLISECONDS);
+							if (!this.locked) {
+								if (acquired) {
+									// Success: we are now leader
+									this.locked = true;
+									handleGranted();
+								}
 							}
 						}
 					}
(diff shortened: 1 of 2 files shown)


pivotal-issuemaster:
@mleguevel Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-integration-extensions&pullRequestId=236)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-integration-extensions&pullRequestId=236) to manually synchronize the status of this Pull Request.

See the [FAQ](https://cla.pivotal.io/about) for frequently asked questions.

pivotal-issuemaster:
@mleguevel Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-integration-extensions&pullRequestId=236)!

artembilan:
Hi, @mleguevel !

Would you mind to explain the fix another way, so it is easy to understand what is going on?
And why have you changed the test?
Don't we need to hon or that Raft algorithm any more?

Thanks

mleguevel:
Hello @artembilan,

**About the fix:**
Currently, hazelcast has a bug when doing an initial `tryLock` on the FencedLock and then doing a lot of `tryLock`/`unlock`. In other words, we can't hold a lock for a long time while also calling `tryLock` and `unlock` frequently on it. This bug causes an internal map to grow indefinitely and thus leading to an OOM. [They have acknowledged](https://github.com/hazelcast/hazelcast/issues/17709#issuecomment-709370454) that it isn't a normal behavior but they don't expect to fix it before version 4.2 (the current version being 4.0.3).
The idea behind this fix is to stop doing the frequent `tryLock` / `unlock`. As I said in the [bug report](https://github.com/spring-projects/spring-integration-extensions/issues/235#issuecomment-708983702), I'm not sure that the algorithm is totally equivalent. I'm not sure about the "else" of the original algorithm in particular:

```java
else {
	this.locked = false;
	// We were not able to acquire it, therefore not leading any more
	handleRevoked();
	if (isRunning()) {
		// Try again quickly in case the lock holder dropped it
		Thread.sleep(LeaderInitiator.this.busyWaitMillis);
	}
}
```
Because I don't know in which case this is invoked. I was hoping you could help me improve the fix if you think that it's not equivalent :)

**About the modifications in the test:**
As I said in the [bug report](https://github.com/spring-projects/spring-integration-extensions/issues/235#issuecomment-708983702), I assumed that you initialized 3 HazelcastInstance in the tests to circumvent the fact that hazelcast 3.12.x does not start in unsafe mode with less than 3 nodes. If my assumptions are not correct, I can set the other 2 instances back.

spring-integration-hazelcast/src/test/java/org/springframework/integration/hazelcast/leader/LeaderInitiatorTests.java
```diff
@@ -227,7 +227,7 @@ public TestCandidate candidate() {
 		@Bean
 		public Config hazelcastConfig() {
 			Config config = new Config();
-			config.getCPSubsystemConfig().setCPMemberCount(3)
+			config.getCPSubsystemConfig().setCPMemberCount(0)
```
artembilan, line 230: Yes, you are right: with HZ-3.12 it "does not start in unsafe mode with less than 3 nodes".
So, this CP Member `0` is something new in HZ 4.0.
Thank you for sharing that and probably it's is good to revise all the tests to work in that unsafe mode when we really don't need a distributed consensus to satisfy 😄

spring-integration-hazelcast/src/main/java/org/springframework/integration/hazelcast/leader/LeaderInitiator.java
```diff
@@ -272,33 +272,19 @@ public Void call() throws Exception {
...
-							}
-						}
-						else {
-							this.locked = false;
-							// We were not able to acquire it, therefore not leading any more
-							handleRevoked();
-							if (isRunning()) {
-								// Try again quickly in case the lock holder dropped it
-								Thread.sleep(LeaderInitiator.this.busyWaitMillis);
+						if (LeaderInitiator.this.lock.isLocked()) {
```
artembilan, line 275: I think this `isLocked()` does the trick for us.
We probably indeed should not re-lock from now on and fully rely on the CP subsystem in HZ.

artembilan (changes requested):
So, I'm merging this with some minor code clean up and adding your name to the author list.

👍

artembilan:
Merged as https://github.com/spring-projects/spring-integration-extensions/commit/b568662cdbd78f6cda084995a2cf0fd7478971ff.

@mleguevel , thank you for contribution; looking forward for more!

Issue #254, opened 2022-03-25
Migrate this repo into Spring Integration repo


ipanimnim:
Opinion - this repo is redundant, confusing, looks unmaintained, and encourages people to not contribute new Spring Integrations Connectors (since these connectors won't enter the Spring Integration project, and instead enter an unknown and undocumented one).

I don't see any advantage to this, and as a new user, I'm just getting more and more confused by this whole thing.

artembilan:
Thanks for bring this up, @ipanimnim .

We already such a discussion in other place: https://github.com/spring-projects/spring-integration/discussions/3720.

Feel free to continue over there.

almogtavor:
@artembilan The discussion we had there was out of subject (the issue was about the request for more extensions in general). This issue seems to fit more. Anyway, was there any progress in this subject since the last discussion?

almogtavor:
@artembilan would you accept PRs that will move this & other repos like `spring-integration-aws` to the main repo?

artembilan:
Not yet. 
We are discussing internally what would be acceptable from this repo.
For example `twitter` is obsolete and based on the library which already out of support. So, it cannot be moved anywhere.
Perhaps to a new repo for obsolete modules a-la `spring-integration-attic`.
The AWS cannot be moved because it is based on Spring Boot and we cannot have circular dependencies.
And so on, so on. Too many questions for every single repo.
For example what do you think about Spring Integration modules in Google GCP project, Microsoft Azure, Alibaba etc?.

At a glance only Cassandra, Hazelcast, Zip, SMB are good to go.

Although I may confirm fully that Goovy DSL is 100% candidate to go into `spring-integration-core` right now.

Thanks

almogtavor:
Why AWS is based on spring boot? About GCP, Alibaba and so, their fine as long as being referenced from the docs. All the others that are relevant should be in the same place, which should encourage to add more connectors over time.

artembilan:
Please, they are not called connectors. They are channel adapters. That's exactly a canonical name from EIP: https://www.enterpriseintegrationpatterns.com/patterns/messaging/ChannelAdapter.html.

The Spring Integration AWS uses Spring Cloud AWS , which is based on Spring Cloud, which, in turn, is based on Spring Boot. And the last one exactly uses Spring Integration. Therefore Spring Integration AWS has to be the last one in the release train.
If we can come up with the solution on a plain AWS SDK (v2?), then it indeed can go to Spring Integration.
[...]

PR #258, opened 2022-04-30
Issue #246 deprecate file replace option, retain minor version

GregBragg:
@artembilan as discussed in PR #257:

"For the current one I suggest the plan like this:
1.	Bring back your SmbMessageHandler, but set replaceFile on the provided SmbSessionFactory, not SmbRemoteFileTemplate.
2.	Deprecate those options on the SmbSessionFactory to discourage end-users to use them explicitly.
3.	Leave the logic as is in the SmbSession as is.

This way we should nail both birds: have options deprecated and retain the old behavior in both write() and rename()."

### DIFF
--- a/spring-integration-smb/README.md
+++ b/spring-integration-smb/README.md
@@ -18,7 +18,7 @@ Put the following block into pom.xml if using Maven:
     <dependency>
         <groupId>org.springframework.integration</groupId>
         <artifactId>spring-integration-smb</artifactId>
-        <version>1.2.0.RELEASE</version>
+        <version>1.3.0.RELEASE</version>
     </dependency>
 
 ## Changes
@@ -30,6 +30,11 @@ Put the following block into pom.xml if using Maven:
 ##### Version 1.2
  * Ability to set the SMB min/max versions in the `SmbSessionFactory` via configuration in the JCIFS library
  * Ability to use a custom implementation of the `jcifs.CIFSContext` interface in the `SmbSessionFactory`
+ 
+##### Version 1.3
+ * Updated to use the latest version of the [JCIFS](https://github.com/codelibs/jcifs) library
+ * Added implementation for `SmbMessageHandler`
+ * Deprecated legacy `replaceFile` and `useTempFile` flags in `SmbConfig`
 
 ## Overview
 
@@ -48,15 +53,14 @@ For XML configuration the `<int-smb:inbound-channel-adapter>` component is provi
 
 ### SMB Outbound Channel Adapter
 
-There is no (yet) some SMB specific requirements for files transferring to SMB, so for XML `<int-smb:outbound-channel-adapter>` component we simply reuse an existing `FileTransferringMessageHandler`.
-In case of Java configuration that `FileTransferringMessageHandler` should be supplied with the `SmbSessionFactory` (or `SmbRemoteFileTemplate`).
+For writing files to a SMB share, and for XML `<int-smb:outbound-channel-adapter>` component we use the `SmbMessageHandler`.
+In case of Java configuration a `SmbMessageHandler` should be supplied with the `SmbSessionFactory` (or `SmbRemoteFileTemplate`).
 
 ````java
-@ServiceActivator(inputChannel = "storeToSmb")
 @Bean
+@ServiceActivator(inputChannel = "storeToSmbShare")
 public MessageHandler smbMessageHandler(SmbSessionFactory smbSessionFactory) {
-    FileTransferringMessageHandler<SmbFile> handler =
-                new FileTransferringMessageHandler<>(smbSessionFactory);
+    SmbMessageHandler handler = new SmbMessageHandler(smbSessionFactory);
     handler.setRemoteDirectoryExpression(
                 new LiteralExpression("remote-target-dir"));
     handler.setFileNameGenerator(m ->
--- a/spring-integration-smb/README.md
+++ b/spring-integration-smb/README.md
@@ -13,12 +13,12 @@ This module adds Spring Integration support for [Server Message Block][] (SMB).
 
 ## Using Maven
 
-Put the following block into pom.xml if using Maven:
+Put the following block into pom.xml if using Maven, set the `<version>` tag to the desired release:
 
     <dependency>
         <groupId>org.springframework.integration</groupId>
         <artifactId>spring-integration-smb</artifactId>
-        <version>1.3.0.RELEASE</version>
+        <version></version>
     </dependency>
 
 ## Changes
@@ -31,7 +31,7 @@ Put the following block into pom.xml if using Maven:
  * Ability to set the SMB min/max versions in the `SmbSessionFactory` via configuration in the JCIFS library
  * Ability to use a custom implementation of the `jcifs.CIFSContext` interface in the `SmbSessionFactory`
  
-##### Version 1.3
+##### Version 1.2.2
  * Updated to use the latest version of the [JCIFS](https://github.com/codelibs/jcifs) library
  * Added implementation for `SmbMessageHandler`
  * Deprecated legacy `replaceFile` and `useTempFile` flags in `SmbConfig`
--- a/spring-integration-smb/src/main/java/org/springframework/integration/smb/outbound/SmbMessageHandler.java
+++ b/spring-integration-smb/src/main/java/org/springframework/integration/smb/outbound/SmbMessageHandler.java
@@ -48,7 +48,6 @@ public class SmbMessageHandler extends FileTransferringMessageHandler<SmbFile> {
 
 	public SmbMessageHandler(SmbRemoteFileTemplate remoteFileTemplate, FileExistsMode mode) {
 		super(remoteFileTemplate, mode);
-		throw new UnsupportedOperationException("do not use, unsupported constructor");
 	}
 
 }
--- a/spring-integration-smb/src/main/java/org/springframework/integration/smb/outbound/SmbMessageHandler.java
+++ b/spring-integration-smb/src/main/java/org/springframework/integration/smb/outbound/SmbMessageHandler.java
@@ -18,8 +18,10 @@ package org.springframework.integration.smb.outbound;
 
 import org.springframework.integration.file.remote.handler.FileTransferringMessageHandler;
 import org.springframework.integration.file.remote.session.SessionFactory;
+import org.springframework.integration.file.support.FileExistsMode;
 import org.springframework.integration.smb.session.SmbRemoteFileTemplate;
 import org.springframework.integration.smb.session.SmbSessionFactory;
+import org.springframework.util.Assert;
 
 import jcifs.smb.SmbFile;
 
@@ -38,6 +40,17 @@ public class SmbMessageHandler extends FileTransferringMessageHandler<SmbFile> {
 	@SuppressWarnings("deprecation")
 	public SmbMessageHandler(SessionFactory<SmbFile> sessionFactory) {
 		super(sessionFactory);
+
+		if (sessionFactory instanceof SmbSessionFactory) {
+			SmbSessionFactory smbSessionFactory = (SmbSessionFactory) sessionFactory;
+			smbSessionFactory.setReplaceFile(true);
+		}
+	}
+
+	@SuppressWarnings("deprecation")
+	public SmbMessageHandler(SessionFactory<SmbFile> sessionFactory, FileExistsMode mode) {
+		super(new SmbRemoteFileTemplate(sessionFactory));
+		Assert.isTrue(mode == FileExistsMode.REPLACE, "Only FileExistsMode.REPLACE is supported.");
 
 		if (sessionFactory instanceof SmbSessionFactory) {
 			SmbSessionFactory smbSessionFactory = (SmbSessionFactory) sessionFactory;
--- a/spring-integration-smb/src/main/java/org/springframework/integration/smb/outbound/SmbMessageHandler.java
+++ b/spring-integration-smb/src/main/java/org/springframework/integration/smb/outbound/SmbMessageHandler.java
@@ -18,7 +18,6 @@ package org.springframework.integration.smb.outbound;
 
 import org.springframework.integration.file.remote.handler.FileTransferringMessageHandler;
 import org.springframework.integration.file.remote.session.SessionFactory;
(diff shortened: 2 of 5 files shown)


spring-integration-smb/src/main/java/org/springframework/integration/smb/outbound/SmbMessageHandler.java
```diff
@@ -0,0 +1,53 @@
...
+		this(new SmbRemoteFileTemplate(sessionFactory));
+		SmbSessionFactory smbSessionFactory = (SmbSessionFactory) sessionFactory;
+		smbSessionFactory.setReplaceFile(true);
+	}
+
+	public SmbMessageHandler(SmbRemoteFileTemplate remoteFileTemplate) {
+		super(remoteFileTemplate);
+	}
+
+	public SmbMessageHandler(SmbRemoteFileTemplate remoteFileTemplate, FileExistsMode mode) {
```
artembilan, line 49: I think we need to disable `SmbRemoteFileTemplate`-based ctors for time being.
Just for the case to not encourage end-users to use those deprecated properties!

spring-integration-smb/src/main/java/org/springframework/integration/smb/outbound/SmbMessageHandler.java
```diff
@@ -0,0 +1,53 @@
...
+* @since 1.3
+*
+* @see SmbRemoteFileTemplate
+*/
+public class SmbMessageHandler extends FileTransferringMessageHandler<SmbFile> {
+
+	public SmbMessageHandler(SessionFactory<SmbFile> sessionFactory) {
+		this(new SmbRemoteFileTemplate(sessionFactory));
+		SmbSessionFactory smbSessionFactory = (SmbSessionFactory) sessionFactory;
+		smbSessionFactory.setReplaceFile(true);
```
artembilan, line 42: I think this one throws a deprecation warning, therefore this ctor has to be marked with the `@SuppressWarning("deprecation")`.
It is totally fine to use a deprecated API in the library code since we need to retain the current behavior.
What we only need to avoid is to let end-users call those deprecated setters from their code: we are going to remove them in the next version anyway.

artembilan (changes requested):

spring-integration-smb/README.md
```diff
@@ -18,7 +18,7 @@ Put the following block into pom.xml if using Maven:
     <dependency>
         <groupId>org.springframework.integration</groupId>
         <artifactId>spring-integration-smb</artifactId>
-        <version>1.2.0.RELEASE</version>
+        <version>1.3.0.RELEASE</version>
```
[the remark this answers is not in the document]
GregBragg: so you want me to remove the documentation on how to add the dependency for a Maven project? Or leave the version as it or change it to something else?

spring-integration-smb/src/main/java/org/springframework/integration/smb/outbound/SmbMessageHandler.java
GregBragg: Not sure I understand... you don't want anybody to use SmbRemoteFileTemplate? I'm just following the pattern for FTP channel adapter in some respects...

spring-integration-smb/README.md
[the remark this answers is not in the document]
artembilan: I think just remove the `<version>` is enough.

spring-integration-smb/src/main/java/org/springframework/integration/smb/outbound/SmbMessageHandler.java
artembilan: That's OK to use it outside of this `SmbMessageHandler`.
The point is that `setReplaceFile(true)` has to be called on the `SmbSessionFactory` and if we expose a `SmbRemoteFileTemplate`-based ctor, that means end-user will have to call that deprecated `setReplaceFile()` manually.
If we don't have this kind of ctors, people will just use an existing one and and won't call a deprecated API, because, yeah, it is deprecated.

Does it make sense?

spring-integration-smb/src/main/java/org/springframework/integration/smb/outbound/SmbMessageHandler.java
```diff
@@ -0,0 +1,53 @@
...
+*
+* @since 1.3
+*
+* @see SmbRemoteFileTemplate
+*/
+public class SmbMessageHandler extends FileTransferringMessageHandler<SmbFile> {
+
+	public SmbMessageHandler(SessionFactory<SmbFile> sessionFactory) {
+		this(new SmbRemoteFileTemplate(sessionFactory));
+		SmbSessionFactory smbSessionFactory = (SmbSessionFactory) sessionFactory;
```
artembilan, line 41: I think we need to have a type check here.
There is a `CachingSessionFactory` which may wrap an original one. So, we will fail here with class cast exception.

artembilan (changes requested):

spring-integration-smb/README.md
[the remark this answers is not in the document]
artembilan: The `gradle.properties` has this content:
```
version=1.2.2.BUILD-SNAPSHOT
```

So, when we release, it is going to be `1.2.2.RELEASE`.

spring-integration-smb/src/main/java/org/springframework/integration/smb/outbound/SmbMessageHandler.java
GregBragg: This is kinda why I wanted to clean up this code to remove the unneeded flags and go to a major version... this is getting messy now. When you see the latest edits you will understand what I mean...

spring-integration-smb/src/main/java/org/springframework/integration/smb/outbound/SmbMessageHandler.java
GregBragg: We are extending `FileTransferringMessageHandler` so I have to support all the constructors in it... How am I going to do that unless I change SmbRemoteFileTemplate like I first proposed in the other branch?

artembilan: Why is that?
There is just enough to have a one ctor based on one super or delegating to another super?
We are not forced to implement all the super ctors.

spring-integration-smb/src/main/java/org/springframework/integration/smb/outbound/SmbMessageHandler.java
artembilan: Yeah...
If you feel so uncomfortable with such a mess, we definitely can look into a new major version.
But that has to go to Spring Integration project already as a module: https://github.com/spring-projects/spring-integration-extensions/issues/254

I just don't want to support two major versions at the same time...

WDYT?

GregBragg: Yeah understood,,, I'm working on trying to get this as clean as possible.

spring-integration-smb/src/test/java/org/springframework/integration/smb/outbound/SmbSendingMessageHandlerTests.java
```diff
@@ -62,7 +62,7 @@ public void prepare() {
 		smbSessionFactory.setUsername("sambaguest");
 		smbSessionFactory.setPassword("sambaguest");
 		smbSessionFactory.setShareAndDir("smb-share/");
-		//smbSessionFactory.setReplaceFile(true);  // deprecated as of 1.3
+		//smbSessionFactory.setReplaceFile(true);  // deprecated as of 1.2.2
```
artembilan, line 65: Why do you still keep this code over here?
I really see it only removed altogether.
Having it commented out does not give us too much value.

artembilan (changes requested):

spring-integration-smb/src/test/java/org/springframework/integration/smb/outbound/SmbSendingMessageHandlerTests.java
GregBragg: removed.

spring-integration-smb/src/main/java/org/springframework/integration/smb/outbound/SmbMessageHandler.java
```diff
@@ -0,0 +1,48 @@
...
+* @author Gregory Bragg
+*
+* @since 1.2.2
+*
+* @see SmbRemoteFileTemplate
+*/
+public class SmbMessageHandler extends FileTransferringMessageHandler<SmbFile> {
+
+	@SuppressWarnings("deprecation")
+	public SmbMessageHandler(SessionFactory<SmbFile> sessionFactory) {
```
artembilan, line 39: I think in addition to this one, I'd like to see this ctor in addition:
```
public SmbMessageHandler(SessionFactory<SmbFile> sessionFactory, FileExistsMode mode) {

```

and do the respective logic depending on the `FileExistsMode`

artembilan (changes requested):

spring-integration-smb/src/main/java/org/springframework/integration/smb/outbound/SmbMessageHandler.java
GregBragg: hoe about this:
	public SmbMessageHandler(SessionFactory<SmbFile> sessionFactory, FileExistsMode mode) {
		this(sessionFactory);
	}

artembilan: Why? Everything is OK. Append, replace, delete etc.
I mean right now we expose only one ctor with the default mode.
To avoid `SmbRemoteFileTemplate` pollution I just suggest to have it based on the `SessionFactory`. Delegate to respective super and call `setReplaceFile()` on the factory when it makes sense based on the provided mode.

GregBragg: Ok I'm obviously misunderstanding here... my apologies... this is a new constructor that doesn't follow any of the base constructors in FileTransferringMessageHandler?

What am I missing?

artembilan: Yes, a new one.
Well, it indeed does not, but it is going to delegate to super based on the built `SmbRemoteFileTemplate`.

Makes sense?

GregBragg: Ok well let me try... not sure I understand the use case fully.

spring-integration-smb/src/main/java/org/springframework/integration/smb/outbound/SmbMessageHandler.java
```diff
@@ -45,4 +47,15 @@ public SmbMessageHandler(SessionFactory<SmbFile> sessionFactory) {
 		}
 	}
 
+	@SuppressWarnings("deprecation")
+	public SmbMessageHandler(SessionFactory<SmbFile> sessionFactory, FileExistsMode mode) {
+		super(new SmbRemoteFileTemplate(sessionFactory));
+		Assert.isTrue(mode == FileExistsMode.REPLACE, "Only FileExistsMode.REPLACE is supported.");
```
artembilan, line 53: Why is that?
The `mode` is not a session factory responsibility.
This is exactly what `FileTransferringMessageHandler` does for us.

What only I would like to see in this ctor is a delegation to the super one like you do now, but with this `mode`  propagation.

In addition it would be a plus to call `setReplaceFile()` if the provided mode is a `FileExistsMode.REPLACE`.

GregBragg: I'm not trying to argue with you but we can't support any mode other than FileExistsMode.REPLACE, why wouldn't we want to inform the developer that only that mode is supported?

artembilan: Not sure in your conclusion. Just follow the `this.remoteFileTemplate.send(message, this.mode)` code and see how that `mode` is used.
There is a logic like in the `RemoteFileTemplate`:
```
	if (FileExistsMode.REPLACE.equals(mode)) {
			session.write(stream, tempFilePath);
		}
		else if (FileExistsMode.APPEND.equals(mode)) {
			session.append(stream, tempFilePath);
		}
```

And this has nothing to do with an SMB protocol. We just call respective `session` method and so on.

GregBragg: I get that... but to fix this bug replaceFile always has to be true. There are a few ways to fix it, only one is backwards compatible and will not drive a major version change.

1. default replaceFile=true in SmbConfig
2. remove all instances of replaceFile code in SmbSession so that the existing file deletes as intended on the SMB share in the write and rename methods
3. default replaceFile=true in all constructors for SmbMessageHandler

All the code for replaceFile and even useTempFile is redundant now, it predates any of this code for FileExistsMode.

I still don't understand what use case/requirement you are trying to solution for here with this new public SmbMessageHandler(SessionFactory<SmbFile> sessionFactory, FileExistsMode mode).

artembilan: My point is to retain the `FileTransferringMessageHandler` functionality as much as possible.
And that one opens for us only if we expose a ctor based on the `FileExistsMode`.
There is nothing special to do for the `SmbSession` since it supports an `APPEND` as is even right now without any changes.

artembilan (approved):
I think your changes are very close to what I would expect as a fix for the issue.

I'll pull it locally today for final review and probably will merge after.

Thank you!

Feel free to look into moving this project as a module to the https://github.com/spring-projects/spring-integration with respective overhaul afterwards for a new major version over there.

artembilan:
Merged as https://github.com/spring-projects/spring-integration-extensions/commit/9ef8e04eade8bfeebcaed3b8d60d90fc4ba260dc after some minor clean ups.

I'm going to release that version today.

Thank you for contribution, @GregBragg ; looking forward for more!

Issue #259, opened 2022-05-13
Restarting HZ cluster will not re-establish any leadership


RobbanHoglund:
I have tested with the code below and it works good when restarting individual HZ members. But when restarting the whole HZ cluster  ( shut down everything), no clients will ever be the leader again (until all clients are restarted).

```
	public static final String LEADER = "leader_lock";
	
	private LeaderInitiator leaderInitiator = null;

	
	@Bean
	public LeaderInitiator initiator(HazelcastInstance instance) {

		return leaderInitiator = new LeaderInitiator(instance, new DefaultCandidate("", LEADER));
	}

	@Scheduled(fixedDelay = 30000 )
	public void runTask() {
		if (leaderInitiator != null) {
			if (leaderInitiator.getContext().isLeader()) {
				LOG.info("I am the leader!");
			} else {
				LOG.info("I am NOT the leader!");
			}

		}
	}
```

artembilan:
I'm not so good with Hazelcast to determine correctly from your description what is HZ cluster and how to configure it.

Any chances to have some simple project from you to reproduce and play with?
Or JUnit test, since I'm good to use a `Hazelcast.newHazelcastInstance()`, therefore I guess something similar should be there for a cluster...

Thanks

RobbanHoglund:
Hi,

Just create a simple springboot application and setup a hazelcast cluster locally in client/server mode (I use version 5.1.1 of Hazelcast).

This is a class/component should work out of the box in simple Springboot applicaiton:

```
import java.util.logging.Logger;

import org.springframework.context.annotation.Bean;
import org.springframework.context.annotation.Configuration;
import org.springframework.integration.dsl.IntegrationFlow;
import org.springframework.integration.dsl.IntegrationFlows;
import org.springframework.integration.hazelcast.leader.LeaderInitiator;
import org.springframework.integration.leader.DefaultCandidate;
import org.springframework.scheduling.annotation.EnableScheduling;
import org.springframework.scheduling.annotation.Scheduled;

import com.hazelcast.core.HazelcastInstance;

@EnableScheduling
@Configuration
public class LeaderElection {
	private static final String CLASS_NAME = LeaderElection.class.getName();

	private static final Logger LOG = Logger.getLogger(CLASS_NAME);

	public static final String LEADER = "myleader";

	private LeaderInitiator leaderInitiator = null;


	public LeaderElection (HazelcastInstance instance) {
	
[...]

PR #260, opened 2022-05-18
GH-259: Ensure lock state after HZ cluster shutdown

RobbanHoglund:
I have tested this implementation/fix against a running HZ cluster (client/server mode) running version 5.1.1.

This will ensure that, if a HZ cluster is totally shut down and then restarted the leader state will be reset if the client still possess/regain the lock.

### DIFF
--- a/spring-integration-hazelcast/src/main/java/org/springframework/integration/hazelcast/leader/LeaderInitiator.java
+++ b/spring-integration-hazelcast/src/main/java/org/springframework/integration/hazelcast/leader/LeaderInitiator.java
@@ -16,12 +16,14 @@
 
 package org.springframework.integration.hazelcast.leader;
 
+
 import java.util.concurrent.Callable;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
 import java.util.concurrent.Semaphore;
 import java.util.concurrent.TimeUnit;
+
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
@@ -39,6 +41,7 @@ import org.springframework.integration.support.leader.LockRegistryLeaderInitiato
 import org.springframework.util.Assert;
 
 import com.hazelcast.core.HazelcastInstance;
+import com.hazelcast.cp.CPSubsystem;
 import com.hazelcast.cp.lock.FencedLock;
 
 /**
@@ -52,6 +55,7 @@ import com.hazelcast.cp.lock.FencedLock;
  * @author Artem Bilan
  * @author Mael Le Guével
  * @author Alexey Tsoy
+ * @author Robert Höglund
  */
 public class LeaderInitiator implements SmartLifecycle, DisposableBean, ApplicationEventPublisherAware {
 
@@ -100,11 +104,6 @@ public class LeaderInitiator implements SmartLifecycle, DisposableBean, Applicat
 	 * This is used to cancel leadership.
 	 */
 	private volatile Future<Void> future;
-
-	/**
-	 * Hazelcast distributed lock.
-	 */
-	private volatile FencedLock lock;
 
 	private boolean customPublisher = false;
 
@@ -210,7 +209,6 @@ public class LeaderInitiator implements SmartLifecycle, DisposableBean, Applicat
 	@Override
 	public synchronized void start() {
 		if (!this.running) {
-			this.lock = this.client.getCPSubsystem().getLock(this.candidate.getRole());
 			this.leaderSelector = new LeaderSelector();
 			this.running = true;
 			this.future = this.executorService.submit(this.leaderSelector);
@@ -252,6 +250,18 @@ public class LeaderInitiator implements SmartLifecycle, DisposableBean, Applicat
 		stop();
 		this.executorService.shutdown();
 	}
+	
+	
+     FencedLock getLock() {
+		CPSubsystem cpSubSystem = this.client.getCPSubsystem();
+		FencedLock lock = cpSubSystem.getLock(this.candidate.getRole());
+		if (logger.isDebugEnabled()) {
+			logger.debug(
+					String.format("Use lock groupid '%s', lock count '%s'", lock.getGroupId(), lock.getLockCount()));
+		}
+		return lock;
+	}
+
 
 	/**
 	 * Callable that manages the acquisition of Hazelcast locks
@@ -263,32 +273,37 @@ public class LeaderInitiator implements SmartLifecycle, DisposableBean, Applicat
 
 		protected final String role = LeaderInitiator.this.candidate.getRole();
 
-		private volatile boolean locked = false;
+		private volatile boolean leader = false;
 
 		@Override
 		public Void call() {
 			try {
 				while (isRunning()) {
 					try {
-						if (LeaderInitiator.this.lock.isLocked()) {
+						if (logger.isTraceEnabled()) {
+							logger.trace("Am I the leader (" + LeaderInitiator.this.candidate.getRole() + ") ? "
+									+ this.leader);
+						}
+						if (getLock().isLockedByCurrentThread()) {
+							if (!this.leader) {
+								// Since we have the lock we need to ensure that the leader flag is set
+								this.leader = true;
+							}
 							// Give it a chance to expire.
 							if (yieldSign.tryAcquire(LeaderInitiator.this.heartBeatMillis, TimeUnit.MILLISECONDS)) {
 								revokeLeadership();
 								// Give it a chance to elect some other leader.
 								Thread.sleep(LeaderInitiator.this.busyWaitMillis);
 							}
-						}
+						} 
 						else {
 							// We try to acquire the lock
-							boolean acquired =
-									LeaderInitiator.this.lock.tryLock(LeaderInitiator.this.heartBeatMillis,
-											TimeUnit.MILLISECONDS);
-							if (!this.locked) {
-								if (acquired) {
+							boolean acquired = getLock()
+									.tryLock(LeaderInitiator.this.heartBeatMillis, TimeUnit.MILLISECONDS);
+							if (acquired && !this.leader) {
 									// Success: we are now leader
-									this.locked = true;
+									this.leader = true;
 									handleGranted();
-								}
 							}
 						}
 					}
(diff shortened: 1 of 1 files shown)


spring-integration-hazelcast/src/main/java/org/springframework/integration/hazelcast/leader/LeaderInitiator.java
```diff
@@ -252,6 +276,18 @@ public void destroy() {
 		stop();
 		this.executorService.shutdown();
 	}
+	
+	
+     FencedLock getLock() {
+		CPSubsystem cpSubSystem = client.getCPSubsystem();
+		FencedLock lock = cpSubSystem.getLock(this.candidate.getRole());
```
artembilan, line 283: So, your point is to get a fresh lock instance every time we would like to acquire it?

spring-integration-hazelcast/src/main/java/org/springframework/integration/hazelcast/leader/LeaderInitiator.java
```diff
@@ -263,30 +299,36 @@ protected class LeaderSelector implements Callable<Void> {
...
 
 		@Override
 		public Void call() {
 			try {
 				while (isRunning()) {
 					try {
-						if (LeaderInitiator.this.lock.isLocked()) {
+						if (logger.isDebugEnabled()) {
+							logger.debug("Am I the leader (" + LeaderInitiator.this.candidate.getRole() + ") ? "
+									+ this.leader);
```
artembilan, line 311: I don't see a reason in this debug message.

artembilan (changes requested):
So, summarizing your change.
It looks like we need only to ensure that on every leader election we call `cpSubSystem.getLock()` instead of that stored into the property.

The rest of your changes are somehow not relevant...

Thanks

spring-integration-hazelcast/src/main/java/org/springframework/integration/hazelcast/leader/LeaderInitiator.java
RobbanHoglund: Yes, or rather get the fresh lock instance every time you want to interact with the lock. according to my experience, caching local refs is not a good practice when it comes to Hazelcast data structures. However the issues I have seen was related to the HZ Failover client (Disaster/Recovery client mode) and when a cluster switch between datacenters happened. But as a practice I always strive to get the data structures from the client every time I want to use them,

spring-integration-hazelcast/src/test/java/org/springframework/integration/hazelcast/leader/LeaderInitiatorTests.java
```diff
@@ -307,5 +310,12 @@ public void publishOnGranted(Object source, Context context, String role) {
 		}
 
 	}
-
+	
+	@Test
+	public void testLeaderElectionLock() throws Exception {
+		LeaderInitiator initiator = new LeaderInitiator(this.hazelcastInstance, new DefaultCandidate());
+		FencedLock lock = initiator.getLock();
+		assertThat(lock).isNotNull();
```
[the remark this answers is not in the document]
RobbanHoglund: Yes the point was to add more value by creating a "FenceLockProvider interface" I could mock. Did not go that far. I will remove it.

spring-integration-hazelcast/src/main/java/org/springframework/integration/hazelcast/leader/LeaderInitiator.java
```diff
@@ -263,30 +299,36 @@ protected class LeaderSelector implements Callable<Void> {
...
+							}
 							// Give it a chance to expire.
 							if (yieldSign.tryAcquire(LeaderInitiator.this.heartBeatMillis, TimeUnit.MILLISECONDS)) {
 								revokeLeadership();
 								// Give it a chance to elect some other leader.
 								Thread.sleep(LeaderInitiator.this.busyWaitMillis);
 							}
-						}
-						else {
+						} else {
```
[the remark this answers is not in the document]
RobbanHoglund: I will fix

spring-integration-hazelcast/src/main/java/org/springframework/integration/hazelcast/leader/LeaderInitiator.java
RobbanHoglund: The reason is to see the thread heartbeat activity when debugging. If it is not needed I can remove it.

RobbanHoglund:
The actual fix is here:
```
if (LeaderInitiator.this.getLock().isLockedByCurrentThread()) {
	if (!this.leader) {
	  // Since we have the lock we need to ensure that the leader flag is set
	 this.leader = true;
	}
```

Btw, Thanks for reviewing my code. :)

spring-integration-hazelcast/src/main/java/org/springframework/integration/hazelcast/leader/LeaderInitiator.java
artembilan: OK. No problem! Then `trace()` might be better 😄

spring-integration-hazelcast/src/main/java/org/springframework/integration/hazelcast/leader/LeaderInitiator.java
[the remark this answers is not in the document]
artembilan: Just for the future contribution: now need to say me if you agree with the change.
I will see that in your subsequent commits anyway.
Otherwise such an `Agreed` causes a noise in my mail box.
And I have like tons of PRs and issue to review/react 😄

spring-integration-hazelcast/src/main/java/org/springframework/integration/hazelcast/leader/LeaderInitiator.java
```diff
@@ -131,7 +130,7 @@ public LeaderInitiator(HazelcastInstance client, Candidate candidate) {
 		this.client = client;
 		this.candidate = candidate;
 	}
-
+
```
artembilan, line 133: Please, make sure there are no trailing whitespaces.
You can run `./gradlew check` to be sure that your changes satisfy our Checkstyle rules.

spring-integration-hazelcast/src/main/java/org/springframework/integration/hazelcast/leader/LeaderInitiator.java
```diff
@@ -252,6 +250,18 @@ public void destroy() {
 		stop();
 		this.executorService.shutdown();
 	}
+	
+	
+     FencedLock getLock() {
+		CPSubsystem cpSubSystem = client.getCPSubsystem();
```
artembilan, line 256: I guess `this.` must be added to that `client` property.

spring-integration-hazelcast/src/main/java/org/springframework/integration/hazelcast/leader/LeaderInitiator.java
```diff
@@ -263,30 +273,37 @@ protected class LeaderSelector implements Callable<Void> {
...
 		public Void call() {
 			try {
 				while (isRunning()) {
 					try {
-						if (LeaderInitiator.this.lock.isLocked()) {
+						if (logger.isTraceEnabled()) {
+							logger.trace("Am I the leader (" + LeaderInitiator.this.candidate.getRole() + ") ? "
+									+ this.leader);
+						}
+						if (LeaderInitiator.this.getLock().isLockedByCurrentThread()) {
```
artembilan, line 287: We don't use prefixed for method calls, unless they are `static`.

spring-integration-hazelcast/src/main/java/org/springframework/integration/hazelcast/leader/LeaderInitiator.java
```diff
@@ -263,30 +273,37 @@ protected class LeaderSelector implements Callable<Void> {
...
 							// We try to acquire the lock
-							boolean acquired =
-									LeaderInitiator.this.lock.tryLock(LeaderInitiator.this.heartBeatMillis,
-											TimeUnit.MILLISECONDS);
-							if (!this.locked) {
-								if (acquired) {
+							boolean acquired = LeaderInitiator.this.getLock()
+									.tryLock(LeaderInitiator.this.heartBeatMillis, TimeUnit.MILLISECONDS);
+							if (acquired) {
+								if (!this.leader) {
```
artembilan, line 304: Looks like these to `ifs` can be combined to a single one.
Do I  miss anything?

spring-integration-hazelcast/src/test/java/org/springframework/integration/hazelcast/leader/LeaderInitiatorTests.java
```diff
@@ -307,5 +306,4 @@ public void publishOnGranted(Object source, Context context, String role) {
 		}
 
 	}
-
```
artembilan, line 310: I think there is no meaningful changes in this class from you.
So, better to revert altogether.

For note: I prefer to have every single member in the class (including the last one, even if it is a nested class) to be wrapped with blank lines. Much easier to read the code.

artembilan (changes requested):

spring-integration-hazelcast/src/main/java/org/springframework/integration/hazelcast/leader/LeaderInitiator.java
RobbanHoglund: No, it is not really needed since we already is in the "this context". But to align with the rest of the code I add it anyway ;)

spring-integration-hazelcast/src/main/java/org/springframework/integration/hazelcast/leader/LeaderInitiator.java
RobbanHoglund: Hmm I am not sure If I follow?

artembilan: It has to be just `getLock().isLockedByCurrentThread()` because this method is not `static` and we use `this.` only for properties.

spring-integration-hazelcast/src/main/java/org/springframework/integration/hazelcast/leader/LeaderInitiator.java
artembilan: We prefer to have it to make it easier to read the code: "is that local variable or the property of the class?"

spring-integration-hazelcast/src/main/java/org/springframework/integration/hazelcast/leader/LeaderInitiator.java
RobbanHoglund: Yes it can, however It was to `ifs` in the original code as well. But sure it can be one if-statement.

artembilan:
Merged as https://github.com/spring-projects/spring-integration-extensions/commit/e11e65e3f542beb76058bc8e277dff6705a59168.

@RobbanHoglund ,

thank you very much for the contribution; looking forward for more!

BTW, we are in agreement to merge this Extensions project with main Spring Integration one: https://github.com/spring-projects/spring-integration-extensions/issues/254.
So, if you wish to contribute more, feel free to migrate HZ module and upgrade it respectively to the latest deps and so!

RobbanHoglund:
Thanks for the good collaboration. And indeed I would gladly contribute if needed, especially with the Hazelcast parts.

artembilan:
> I would gladly contribute if needed

Right, as I said: we need now to migrate HZ extension project to Spring Integration as a module.
We did recently something similar for SMB one: https://github.com/spring-projects/spring-integration/pull/3791.

So, that's kinda the story we may consider as a contribution at the moment for HZ.

Thanks

RobbanHoglund:
Hi, 
I did an attempt here:
https://github.com/spring-projects/spring-integration/pull/3812

